PR #78, opened 2022-03-15
Create FuncOp for direct call if callee is not available

kumarak:


xlauko:
Fixes merged in other commits

PR #183, opened 2022-10-31
Lower hl func op

lkorenc:
Add lowering of `hl::FuncOp` into `func::FuncOp` which can be then lowered further using standard mlir passes.
Add a few fixes of transformation to LLVM.
Enable previously disabled tests.


lib/vast/Dialect/HighLevel/Transforms/HLLowerTypes.cpp
```diff
@@ -518,6 +518,7 @@ namespace vast::hl
             FuncOp new_fn = rewriter.create< FuncOp >(
                 fn.getLoc(), fn.getName(), fn_type, fn.getLinkage(), attributes
             );
+            new_fn.setVisibility(mlir::SymbolTable::Visibility::Private);
```
xlauko, line 521: can not you inherit visibility from `fn`?

xlauko (approved):

PR #184, opened 2022-10-31
Fix type assertion due to invalid cast

### LINKED ISSUE #171: Type Assertion due to invalid cast
I see type assertion due to invalid cast from NamedType to FunctionType. It happens because typedef is getting lowered as NamedType. The issue can be reproduce with the code:

```
typedef int function_type(int a, int b);
function_type *p[2];
int test_func(void) {
  return p[1](15, 20);
}
```
https://github.com/trailofbits/vast/blob/master/lib/vast/Dialect/HighLevel/HighLevelTypes.cpp#L18
https://github.com/trailofbits/vast/blob/master/lib/vast/Dialect/HighLevel/HighLevelTypes.cpp#L23

xlauko:
- implements `getFunctionType` for typedef of function pointers. 
- updates `IndirectCallOp` to allow typedef `callee` type.

closes #171

### DIFF
--- a/include/vast/Dialect/HighLevel/HighLevelTypes.hpp
+++ b/include/vast/Dialect/HighLevel/HighLevelTypes.hpp
@@ -13,9 +13,11 @@ VAST_RELAX_WARNINGS
 #include <mlir/IR/MLIRContext.h>
 #include <mlir/IR/TypeSupport.h>
 #include <mlir/IR/Types.h>
+#include <mlir/Interfaces/CallInterfaces.h>
 #include <mlir/Interfaces/DataLayoutInterfaces.h>
 VAST_UNRELAX_WARNINGS
 
+#include "vast/Util/Common.hpp"
 #include "vast/Util/DataLayout.hpp"
 #include "vast/Util/TypeList.hpp"
 #include "vast/Util/Types.hpp"
@@ -157,8 +159,16 @@ namespace vast::hl
         }
     }
 
-    mlir::FunctionType getFunctionType(PointerType functionPointer);
-    mlir::FunctionType getFunctionType(mlir::Type functionPointer);
+    mlir::FunctionType getFunctionType(Type function_pointer, Module mod);
+
+    mlir::FunctionType getFunctionType(Value callee);
+    mlir::FunctionType getFunctionType(mlir::CallOpInterface call);
+    mlir::FunctionType getFunctionType(mlir::CallInterfaceCallable callee, Module mod);
+
+    Type getTypedefType(TypedefType type, Module mod);
+
+    // unwraps all typedef aliases to get to real underlying type
+    Type getBottomTypedefType(TypedefType def, Module mod);
 
     bool isBoolType(mlir::Type type);
     bool isIntegerType(mlir::Type type);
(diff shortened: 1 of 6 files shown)


include/vast/Dialect/HighLevel/HighLevelTypes.hpp
```diff
@@ -157,8 +159,15 @@ namespace vast::hl
...
-    mlir::FunctionType getFunctionType(mlir::Type functionPointer);
+    mlir::FunctionType getFunctionType(Type function_pointer, Module mod);
+
+    mlir::FunctionType getFunctionType(Value callee);
+    mlir::FunctionType getFunctionType(mlir::CallOpInterface call);
+    mlir::FunctionType getFunctionType(mlir::CallInterfaceCallable callee, Module mod);
+    mlir::FunctionType getFunctionType(mlir::CallInterfaceCallable callee, Module mod);
+
+    Type getTypedefType(TypedefType type, Module mod);
+    Type getBottomTypedefType(TypedefType def, Module mod);
```
lkorenc, line 170: What does `Bottom` mean here? The first type that is not typedef anymore? Add doc.

include/vast/Dialect/HighLevel/HighLevelTypes.hpp
```diff
@@ -157,8 +159,15 @@ namespace vast::hl
...
     }
 
-    mlir::FunctionType getFunctionType(PointerType functionPointer);
-    mlir::FunctionType getFunctionType(mlir::Type functionPointer);
+    mlir::FunctionType getFunctionType(Type function_pointer, Module mod);
+
+    mlir::FunctionType getFunctionType(Value callee);
+    mlir::FunctionType getFunctionType(mlir::CallOpInterface call);
+    mlir::FunctionType getFunctionType(mlir::CallInterfaceCallable callee, Module mod);
+    mlir::FunctionType getFunctionType(mlir::CallInterfaceCallable callee, Module mod);
```
lkorenc, line 167:
```suggestion
```

Looks like copy and paste error?

lib/vast/Dialect/HighLevel/HighLevelTypes.cpp
```diff
@@ -29,19 +30,70 @@ namespace vast::hl
...
-            return getFunctionType(ty.getElementType());
+            return getFunctionType(ty.getElementType(), mod);
         if (auto ty = type.dyn_cast< PointerType >())
-            return getFunctionType(ty.getElementType());
+            return getFunctionType(ty.getElementType(), mod);
+        if (auto ty = type.dyn_cast< TypedefType >()) {
+            return getFunctionType(getTypedefType(ty, mod), mod);
+        }
+
+        type.dump();
```
lkorenc, line 67:
```suggestion
```

Forgotten debug print? Or is that supposed to be part of report if `UNREACHABLE` is reached?

lkorenc (approved):

include/vast/Dialect/HighLevel/HighLevelTypes.hpp
xlauko: Yes it unwraps all typedefs, do you have a better name?

lkorenc: Nope, I guess just add a comment in case someone is confused when reading it?

PR #203, opened 2022-12-05
Add pattern rewriter mixin.

xlauko:


include/vast/Conversion/Common/Patterns.hpp
```diff
@@ -12,10 +12,52 @@ namespace vast {
...
     using pattern_rewriter = mlir::PatternRewriter;
 
+    using conversion_rewriter = mlir::ConversionPatternRewriter;
+
+    template< typename derived_pattern >
+    struct pattern_mixin {
+
+        using integer_attr_t = mlir::IntegerAttr;
+
+        derived_pattern& self() { return *static_cast< derived_pattern* >(this); }
```
lkorenc, line 22:
```suggestion
        derived_pattern& self() { return static_cast< derived_pattern &>(*this); }
```

include/vast/Conversion/Common/Patterns.hpp
```diff
@@ -12,10 +12,52 @@ namespace vast {
...
+        const derived_pattern& self() const { return *static_cast< const derived_pattern* >(this); }
```
lkorenc, line 23:
```suggestion
        const derived_pattern& self() const { return static_cast< const derived_pattern &>(*this); }
```

include/vast/Conversion/Common/Patterns.hpp
```diff
@@ -12,10 +12,52 @@ namespace vast {
...
+
+        template< typename T >
+        mlir_type bitwidth_type() const {
+            auto ctx = self().getContext();
+            return mlir::IntegerType::get(ctx, bits< T >());
```
lkorenc, line 28: This is kinda "locking" user into mlir dialect - which is not bad by itself, but maybe we want to reflect that in the mixin name?

lkorenc (approved):

PR #220, opened 2023-01-21
Null check for type pointer before calling Visitor

### LINKED ISSUE #218: Assert failure on getting the Type pointer
The underlying type pointer of a `QualType` can be null. Function `getTypePtr` will assert in that case. Use `getTypePtrOrNull` instead and handle if the type pointer is null.

```
#3  0x00007ffff7a76fd6 in __GI___assert_fail (assertion=0x55555c1b0025 "!isNull() && \"Cannot retrieve a NULL type pointer\"", 
    file=0x55555c1afe98 "/home/user/workspace/vcpkg-llvm15-install/installed/x64-linux-rel/include/clang/AST/Type.h", line=747, 
    function=0x55555c1b0058 "const clang::ExtQualsTypeCommonBase *clang::QualType::getCommonPtr() const") at assert.c:101
#4  0x000055555608f037 in clang::QualType::getCommonPtr() const ()
#5  0x0000555556090389 in clang::QualType::getQualifiers() const ()
#6  0x000055555608dd54 in vast::hl::CodeGenTypeVisitorMixin<vast::hl::CodeGenVisitor<indexer::CodeGeneratorVisitor::VisitorConfig, indexer::MetaGenerator> >::Visit(clang::QualType) ()
#7  0x000055555608daf5 in vast::hl::CodeGenTypeVisitorWithDataLayoutMixin<vast::hl::CodeGenVisitor<indexer::CodeGeneratorVisitor::VisitorConfig, indexer::MetaGenerator> >::Visit(clang::QualType) ()

```

kumarak:
The PR adds check for the null underlying type pointer before calling the type visitor and storing the data layout. If the type pointer is null, return empty `mlir::Type` object. (Fixes #218)

### DIFF
--- a/include/vast/Translation/CodeGenTypeVisitor.hpp
+++ b/include/vast/Translation/CodeGenTypeVisitor.hpp
@@ -136,7 +136,9 @@ namespace vast::hl {
         }
 
         auto Visit(clang::QualType ty) -> mlir_type {
-            auto underlying = ty.getTypePtr();
+            auto underlying = ty.getTypePtrOrNull();
+            VAST_ASSERT(underlying != nullptr);
+
             auto quals      = ty.getQualifiers();
             if (auto t = llvm::dyn_cast< clang::BuiltinType >(underlying)) {
                 return VisitBuiltinType(t, quals);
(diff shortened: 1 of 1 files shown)


xlauko:
When such thing happens? Can you create an example and add it to test-suite?

lkorenc:
Bumping this.

PR #228, opened 2023-02-08
Add conversion utils for passes to llvm

Jezurko:


include/vast/Conversion/Common/Passes.hpp
```diff
@@ -89,4 +91,79 @@ namespace vast {
...
+            if constexpr ( list::empty ) {
+                return;
+            } else {
+                patterns.add< typename list::head >(converter);
+                return populate_conversions_base< typename list::tail >(patterns, converter);
+            }
+        }
+
+        template< typename ...lists  >
+        static void populate_conversions_base(rewrite_pattern_set &patterns, type_converter &converter) {
```
lkorenc, line 143: Make whatever method needs to be overwritten `virtual`.

include/vast/Conversion/Common/Patterns.hpp
```diff
@@ -61,4 +62,43 @@ namespace vast {
...
+
+        auto fN(auto &rewriter, auto loc, mlir::Type type, auto val) const
+        {
+            return rewriter.template create< mlir::LLVM::ConstantOp >(
+                    loc,
+                    type,
+                    rewriter.getFloatAttr(type, val));
+        }
+
+        auto anyN(auto &rewriter, auto loc, mlir::Type type, auto val) const
```
lkorenc, line 83: `constant(`?

lkorenc (approved):

include/vast/Conversion/Common/Passes.hpp
Jezurko: Right now no methods need to be overwritten. In 6915f1d901be8c4182ba13e26d7d95f7ff572415 I modified it so that the method implemented by the user is `populate_conversions` and the method in the base class is `populate_conversions_base` (so that it explodes if there is no `populate_conversions` method). Should I rename it back?

lkorenc: Sorry, I missed that! This is fine.

PR #236, opened 2023-03-01
Control flow lowering

lkorenc:
- Remove pass that lowered control flow into `SCF`.
  - Remove associated tests.
- Remove pass that lowered `hl::FuncOp` into `func::`. 
- Add pass that lowers control flow into `ll`.
  - Add appropriate operations to `ll`.
- Add transformation that removes dead code in `hl` (for example operations after first `return`).
- Add draft of `Target` component.


lib/vast/Conversion/Common/Common.hpp
```diff
@@ -0,0 +1,119 @@
...
+        auto new_bw = dl.getTypeSizeInBits(target);
+        auto original_bw = dl.getTypeSizeInBits(op.getType());
+
+        if (new_bw == original_bw)
+            return op;
+        else if (new_bw > original_bw)
+            return rewriter.template create< mlir::LLVM::SExtOp >(loc, target, op);
+        else
+            return rewriter.template create< mlir::LLVM::TruncOp >(loc, target, op);
+    }
```
Jezurko, line 72: Maybe we can move it to `Conversion/Common/Patterns.hpp` to the `llvm_pattern_utils` struct?

lib/vast/Conversion/Common/IRsToLLVM.cpp
```diff
@@ -763,10 +695,16 @@ namespace vast
...
-            target.addDynamicallyLegalOp< hl::ValueYieldOp >(has_llvm_return_type< hl::ValueYieldOp >);
+            auto illegal_with_llvm_ret_type = [&]< typename T >( T && )
+            {
+                target.addDynamicallyLegalOp< T >( has_llvm_return_type< T > );
+            };
+
+            illegal_with_llvm_ret_type( core::LazyOp{} );
+            illegal_with_llvm_ret_type( core::BinLAndOp{} );
+            illegal_with_llvm_ret_type( core::BinLOrOp{} );
+            illegal_with_llvm_ret_type( hl::ValueYieldOp{} );
```
Jezurko, line 706: Why `illegal`? Shouldn't it be `legal_with_llvm_ret_type`? We are allowing the op only if the return type is in llvm.
```suggestion
            auto legal_with_llvm_ret_type = [&]< typename T >( T && )
            {
                target.addDynamicallyLegalOp< T >( has_llvm_return_type< T > );
            };

            legal_with_llvm_ret_type( core::LazyOp{} );
            legal_with_llvm_ret_type( core::BinLAndOp{} );
            legal_with_llvm_ret_type( core::BinLOrOp{} );
            legal_with_llvm_ret_type( hl::ValueYieldOp{} );
```

include/vast/Dialect/LowLevel/LowLevelOps.td
```diff
@@ -39,4 +41,138 @@ def InitializeVar
         to constructors.
     }];
 }
+
+def Br
```
xlauko, line 45: I did not have time to check yet, but I think we overlap in some parts here with core/high-level dialects.

For example we may want to have just single scoping primitives, e.g., scope end as `yield` in code?

lib/vast/Conversion/Common/Common.hpp
lkorenc: Yeah that is probably a good idea.

include/vast/Conversion/Common/Passes.hpp
```diff
@@ -127,21 +179,33 @@ namespace vast {
...
-                return populate_conversions_base< typename list::tail >(patterns, converter);
-            }
-        }
-
-        template< typename ...lists  >
-        static void populate_conversions_base(rewrite_pattern_set &patterns, type_converter &converter) {
-            (populate_conversions_impl< lists >(patterns, converter), ...);
+            rewrite_pattern_set patterns;
+            conversion_target target;
+            // Type converter cannot be moved!
```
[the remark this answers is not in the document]
lkorenc: I am honestly not sure, will dig into sources.

include/vast/Util/LLVMTypeConverter.hpp
```diff
@@ -142,8 +155,12 @@ namespace vast::util::tc
                 return {};
 
             // LLVM function can have only one return value;
-            if (r_res->size() != 1)
+            if (r_res->size() > 1)
```
xlauko, line 158: shouldn't this be assert?

include/vast/Dialect/LowLevel/LowLevelOps.td
lkorenc: Will open an issue and merge this in separate PR.

lib/vast/Target/LLVMIR/Convert.cpp
```diff
@@ -0,0 +1,97 @@
...
+    class ToLLLVMIR : public mlir::LLVMTranslationDialectInterface
+    {
+      public:
+        using Base = mlir::LLVMTranslationDialectInterface;
+        using Base::Base;
+
+        mlir::LogicalResult convertOperation(mlir::Operation *op, llvm::IRBuilderBase &irb,
+                                             mlir::LLVM::ModuleTranslation &state) const final
+        {
+            return llvm::TypeSwitch< mlir::Operation *, mlir::LogicalResult >(op)
```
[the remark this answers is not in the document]
lkorenc: For now yeah, but eventually there will most likely be some hooks that do something, so I am not collapsing it yet.

xlauko (approved):

PR #241, opened 2023-03-07
Update linking library to LLVMTarget

kumarak:
Update vast-front linking library to use LLVMTarget instead of linking to LLVM. This is to fix build error with vcpkg.


kumarak:
Hi @Jezurko, Could you check if it works with your local build of llvm?

xlauko (approved):

kumarak:
I will go ahead and merge it. @xlauko, Thanks for updating the PR.

PR #251, opened 2023-03-15
Refactor CF region-related code

xlauko:


lib/vast/Dialect/HighLevel/HighLevelOps.cpp
```diff
@@ -24,15 +25,7 @@ namespace vast::hl
...
-        {
-            auto reg = st.addRegion();
-            if (callback.has_value()) {
-                bld.createBlock(reg);
-                callback.value()(bld, st.location);
-            }
-            return reg;
-        }
+
     } // namespace detail
```
Jezurko, line 29: Is there a point in keeping the empty namespace?

lib/vast/Dialect/HighLevel/HighLevelOps.cpp
```diff
@@ -201,15 +194,15 @@ namespace vast::hl
...
 
     namespace detail {
         void build_record_like_decl(Builder &bld, State &st, llvm::StringRef name, BuilderCallback fields) {
             st.addAttribute("name", bld.getStringAttr(name));
 
             InsertionGuard guard(bld);
-            detail::build_region(bld, st, fields);
+            build_region(bld, st, fields);
         }
     } // anamespace detail
```
Jezurko, line 207: Ah, ok. It's not empty.

Jezurko (approved):

PR #258, opened 2023-03-17
Extract types from `LValue` in `CondOp` verification

Jezurko:


lib/vast/Dialect/HighLevel/HighLevelOps.cpp
```diff
@@ -297,9 +297,17 @@ namespace vast::hl
...
+        auto normalize = [](mlir::Type type) {
+           auto type_casted = type.dyn_cast< hl::LValueType >();
+           if (type_casted)
+               return type_casted.getElementType();
+            return type;
+        };
+        auto lhs_norm = normalize(lhs);
+        auto rhs_norm = normalize(rhs);
+        bool compatible = lhs_norm == rhs_norm
+            || lhs_norm.hasTrait< tt::TypedefTrait >() || rhs_norm.hasTrait< tt::TypedefTrait >()
```
xlauko, line 309: I would add trait utilities to write something similar to:
```
compatible = (l == r) || any_of< has_trait< tt::TypedefTrait >(l, r) || all_of< has_trait< tt::PointerTypeTrait > >(l, r);
``` 
```

File: lib/vast/Dialect/HighLevel/HighLevelOps.cpp [hunk 2]
```diff
@@ -297,9 +297,17 @@ namespace vast::hl
...
+            || (lhs_norm.hasTrait< tt::PointerTypeTrait >() && rhs_norm.hasTrait< tt::PointerTypeTrait >());
         VAST_ASSERT(compatible && "failed to verify CondOp types");
```
xlauko, line 311: why is this assert here? Isn't this checked and responsibility of the caller of `areTypesCompatible`?

Jezurko: The assert is there to print at least some error message in case the verification fails. Without it `vast-cc` just ends silently returning `1`. I should change it to something less destructive probably?

lib/vast/Dialect/HighLevel/HighLevelOps.cpp
```diff
@@ -273,42 +274,33 @@ namespace vast::hl
...
 
     logical_result CondOp::verifyRegions()
     {
         auto then_type = get_yielded_type(getThenRegion());
         auto else_type = get_yielded_type(getElseRegion());
-        return mlir::success(areTypesCompatible(then_type, else_type));
+        bool compatible = typesMatch(then_type, else_type);
+        if (!compatible)
+        {
+            llvm::errs() << "Failed to verify that return types " << then_type
```
kumarak, line 300: I do see a few times this condition hits and two region types does not match. This is what I see in the log if that helps:
```
Failed to verify that return types !hl.int, Failed to verify that return types !!hlhl..long< unsigned >, long< unsigned >! in CondOp regions match.
hl.int in CondOp regions match.
Failed to verify that return types !hl.int, !hl.long< unsigned > in CondOp regions match.
Failed to verify that return types !hl.long< unsigned >, !hl.int in CondOp regions match.
```

xlauko: @kumarak can you investigate this more, it seams like invalid AST to me. There need to be implicit integer cast from `hl.int` to `hl.long` somewhere.

Jezurko:
@kumarak Did you have time to find an example/snippet that is causing the issue? I would like to have a test case for this before we merge it

kumarak:
@Jezurko, I will get you the AST or generated MLIR that was causing issue today. That could be helpful for writing test cases.

lib/vast/Dialect/HighLevel/HighLevelOps.cpp
kumarak: The error logs was due to not operation and got fixed with the PR #264.

lkorenc:
Should be already merged in master via other PRs.

PR #264, opened 2023-03-23
Fix return type of high-level LNotOp

### LINKED ISSUE #257: Type mismatch for CondOp regions
Assertion due to type mismatch of `then` and `else` yields for conditional operation. The operation verification fails.

```
[vast] assertion: compatible && "failed to verify CondOp types" failed
UNREACHABLE executed at /Users/akkumar/Documents/workspace/src/vast/lib/vast/Dialect/HighLevel/HighLevelOps.cpp:306!
```
This is the `then` and `else` type getting yield on dumping the types.

then region: `!hl.lvalue<!hl.ptr<!hl.char>>`
else region: `!hl.ptr<!hl.char>`

Jezurko:
This should fix #257. @kumarak could you please confirm?

### DIFF
--- a/include/vast/Dialect/HighLevel/HighLevelOps.td
+++ b/include/vast/Dialect/HighLevel/HighLevelOps.td
@@ -986,7 +986,26 @@ class TypePreservingUnOp< string mnemonic, list< Trait > traits = [] >
 def PlusOp  : TypePreservingUnOp< "plus", [] >;
 def MinusOp : TypePreservingUnOp< "minus", [] >;
 def NotOp   : TypePreservingUnOp< "not", [] >;
-def LNotOp  : TypePreservingUnOp< "lnot", [] >;
+
+class LogicalUnOp< string mnemonic, list< Trait > traits = [] >
+    : HighLevel_Op< mnemonic, !listconcat(traits, [NoSideEffect]) >
+    , Arguments< (ins AnyType:$arg) >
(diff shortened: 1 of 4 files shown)


include/vast/Dialect/HighLevel/HighLevelOps.td
```diff
@@ -986,7 +986,26 @@ class TypePreservingUnOp< string mnemonic, list< Trait > traits = [] >
 def PlusOp  : TypePreservingUnOp< "plus", [] >;
 def MinusOp : TypePreservingUnOp< "minus", [] >;
 def NotOp   : TypePreservingUnOp< "not", [] >;
-def LNotOp  : TypePreservingUnOp< "lnot", [] >;
+
+class LogicalUnOp< string mnemonic, list< Trait > traits = [] >
+    : HighLevel_Op< mnemonic, !listconcat(traits, [NoSideEffect]) >
+    , Arguments< (ins AnyType:$arg) >
```
xlauko, line 992: I did not check the standard but isn't this also required to be int or bool?

xlauko (approved):

include/vast/Dialect/HighLevel/HighLevelOps.td
Jezurko: C++ does require conversion of the operand to bool but the C standard only says that the operand is a scalar type :/

PR #266, opened 2023-03-24
cg: Deduplicate functions using their mangled name.

### LINKED ISSUE #265: Function redeclaration error
Deal with compilation of:

```
void abort();
void abort();
int main() {}
```

respectively:

```
void abort();
void abort() {}
int main() {}
```

This is minimal example of #253

xlauko:
- tries to deduplicate function definitions using their mangled names

- fixes #265


xlauko:
At the moment I see nondeterministically inbalanced scope errors, don't knwo how to solve it yet.

PR #273, opened 2023-04-03
Frontend action to emit llvm/mlir and associated fixes

lkorenc:
Add support to `vast-front` to emit llvm IR (`-vast-emit-llvm`) or some selected mlir dialects (`-vast-emit-mlir=[hl,llvm]`).

Fixes
 * Bug in control flow lowering that improperly routed `hl.continue` to `init` instead of `incr`.
 * Ternary operator in case its operands are `void`.

### DIFF
--- a/lib/vast/Dialect/HighLevel/HighLevelOps.cpp
+++ b/lib/vast/Dialect/HighLevel/HighLevelOps.cpp
@@ -297,6 +297,12 @@ namespace vast::hl
     bool CondOp::typesMatch(mlir::Type lhs, mlir::Type rhs)
     {
         namespace tt = mlir::TypeTrait;
+
+        if (!lhs)
+            return !rhs || rhs.isa< hl::VoidType, mlir::NoneType >();
+        if (!rhs)
+            return !lhs || lhs.isa< hl::VoidType, mlir::NoneType >();
+
         return lhs == rhs
             || all_with_trait< tt::IntegralTypeTrait >(lhs, rhs)
             || any_with_trait< tt::TypedefTrait >(lhs, rhs)
@@ -307,10 +313,6 @@ namespace vast::hl
     {
         auto then_type = get_maybe_yielded_type(getThenRegion());
         auto else_type = get_maybe_yielded_type(getElseRegion());
-        if ( !then_type )
-            return mlir::success(!else_type || else_type.isa< hl::VoidType >());
-        if ( !else_type )
-            return mlir::success(!then_type || then_type.isa< hl::VoidType >());
 
         bool compatible = typesMatch(then_type, else_type);
         if (!compatible)
--- a/lib/vast/Dialect/HighLevel/HighLevelOps.cpp
+++ b/lib/vast/Dialect/HighLevel/HighLevelOps.cpp
@@ -305,8 +305,13 @@ namespace vast::hl
 
     logical_result CondOp::verifyRegions()
     {
-        auto then_type = get_yielded_type(getThenRegion());
-        auto else_type = get_yielded_type(getElseRegion());
+        auto then_type = get_maybe_yielded_type(getThenRegion());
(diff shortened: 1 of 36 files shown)


lib/vast/Dialect/HighLevel/HighLevelOps.cpp
```diff
@@ -305,8 +305,13 @@ namespace vast::hl
...
     {
-        auto then_type = get_yielded_type(getThenRegion());
-        auto else_type = get_yielded_type(getElseRegion());
+        auto then_type = get_maybe_yielded_type(getThenRegion());
+        auto else_type = get_maybe_yielded_type(getElseRegion());
+        if ( !then_type )
+            return mlir::success(!else_type || else_type.isa< hl::VoidType >());
+        if ( !else_type )
+            return mlir::success(!then_type || then_type.isa< hl::VoidType >());
+
```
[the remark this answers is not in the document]
Jezurko:
```suggestion
        if ( !then_type )
            return mlir::success( !else_type || else_type.isa< hl::VoidType >() );
        if ( !else_type )
            return mlir::success( !then_type || then_type.isa< hl::VoidType >() );

```

Jezurko (approved):

lib/vast/Frontend/GenAction.cpp
```diff
@@ -43,6 +53,7 @@ namespace vast::cc {
             case output_type::emit_assembly: return "s";
             case output_type::emit_high_level: return "hl";
             case output_type::emit_cir: return "cir";
+            case output_type::emit_mlir: return "mlir";
```
xlauko, line 56: either keep only `emit_mlir` or targets for specific dialects, otherwise it will be mess

lib/vast/Frontend/GenAction.cpp
```diff
@@ -132,33 +145,88 @@ namespace vast::cc {
...
+            if (list->size() != 1)
+                VAST_UNREACHABLE("Can emit only one dialect.");
+
+            return parse_target_dialect(list->front());
+        }
+
+        target_dialect parse_target_dialect(llvm::StringRef from)
+        {
+            auto trg = from.lower();
+            if (trg == "hl" || trg == "high_level")
```
xlauko, line 169: we should add some single place description of dialects and use named constants or some other tag-dispatch-like mechanism, now it is starting to be bunch or random string all over the place

lib/vast/Frontend/GenAction.cpp
```diff
@@ -113,12 +124,14 @@ namespace vast::cc {
         void HandleInterestingDecl(clang::DeclGroupRef /* decl */) override {
             VAST_UNIMPLEMENTED;
         }
+        void emit_backend_output(clang::BackendAction backend_action,
+                                 owning_module_ref mlir_module, mcontext_t *mctx)
+        {
+            llvm::LLVMContext llvm_context;
+            target::llvmir::register_vast_to_llvm_ir(*mctx);
+            target::llvmir::prepare_hl_module(mlir_module.get());
```
xlauko, line 132: this has nothing with `hl` module prepare. Rename to something like `setup_lowering_pipeline`

xlauko: idealy we create single entity which will take `mlir_target` object and will setup required passes

lib/vast/Frontend/GenAction.cpp
```diff
@@ -132,33 +145,88 @@ namespace vast::cc {
...
+        {
+            switch (target)
+            {
+                case target_dialect::high_level: return "high_level";
+                case target_dialect::low_level: return "low_level";
+                case target_dialect::llvm: return "llvm";
+            }
+        }
+
+        void compile_via_vast(auto mod, mcontext_t *mctx)
```
xlauko, line 188: if we are in `vast_gen_action` we always compile via vast

PR #276, opened 2023-04-07
cmake: add MLIRLowLevel to vast install targets

kumarak:


lkorenc:
For future: commit message should start with capital (after `component:`) and end with dot.

lkorenc (approved):

PR #316, opened 2023-06-21
Initial support for C++

frabert:
Missing in this PR:
  - `friend` declarations
  - Separate operations for methods, constructors and destructors (with support for the various attributes that those declarations have, like `virtual`, `volatile`, `const`, `&`, `&&`...)
  - Constructor initializers


include/vast/Translation/CodeGenContext.hpp
```diff
@@ -130,7 +130,7 @@ namespace vast::cg
             return "anonymous[" + std::to_string(decl->getID()) + "]";
         }
 
-        std::string get_namespaced_for_decl_name(const clang::TagDecl *decl) {
+        std::string get_namespaced_for_decl_name(const clang::NamedDecl *decl) {
```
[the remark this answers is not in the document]
frabert: `VarDecl`s also need to have a scoped name in the case of static variables inside records (I think).
`VarDecl`s are not `TagDecl`s, but both are `NamedDecl`s so this allows using the same logic for both cases

lkorenc:
Also please move the newly added test to `HighLevel/CXX` (so it kind of mirrors the namespace `hl:cxx`).

include/vast/Dialect/HighLevel/HighLevelTypes.td
```diff
@@ -272,6 +272,21 @@ class TypesOrLValueTypesMatch<string summary, string lhs, string rhs> : PredOpTr
   ]>
 >;
 
+//
+// RValue high-level type
+//
+def RValue : HighLevelType< "RValue",
```
lkorenc, line 278: Copy & paste of `lvalue` extract that to shared helper.

include/vast/Dialect/HighLevel/HighLevelTypes.td
```diff
@@ -442,4 +457,10 @@ def AdjustedType : HighLevelType< "Adjusted" > {
   let assemblyFormat = "`<` $original `,` $adjusted `>`";
 }
 
+def ReferenceType : HighLevelType< "Reference" > {
```
lkorenc, line 460: Can use the same helper as lvalue/rvalue. Also should implement `SubElementTypeInterface`.

lkorenc (changes requested):

lkorenc (approved):

PR #317, opened 2023-06-26
Fixup `goto`'s in `vast-front`

Jezurko:
- Move splicing of trailing scopes to a separate pass
- Emit `LabelDecl`s as the first think in function generation

### DIFF
--- a/include/vast/Translation/CodeGenDeclVisitor.hpp
+++ b/include/vast/Translation/CodeGenDeclVisitor.hpp
@@ -354,22 +354,25 @@ namespace vast::cg {
                 // Making sure, that if the operation is enclosed in a trailing
                 // scope, then the termiantor is evaluated in this scope (which
                 // will then be spliced by subsequent pass)
-                auto scope = mlir::dyn_cast< hl::ScopeOp >(ops.back());
+                auto next_scope = [](operation op) -> hl::ScopeOp {
+                    if (op)
+                        return mlir::dyn_cast< hl::ScopeOp >(op);
+                    return {};
+                };
 
-                while (scope) {
+                auto process_scope = [&](hl::ScopeOp scope) -> operation {
                     auto parent = scope->getParentRegion();
                     if (parent->hasOneBlock()
                         && parent->back().begin() == std::prev(parent->back().end()))
                     {
-                        last_op = get_last_op(scope);
                         set_insertion_point_to_end(&scope.getBody());
-                        if (last_op)
-                            scope = mlir::dyn_cast< hl::ScopeOp >(last_op);
-                        else
-                            scope = nullptr;
-                    } else {
-                        break;
+                        return get_last_op(scope);
                     }
+                    return {};
+                };
+
+                while (auto scope = next_scope(last_op)) {
+                    last_op = process_scope(scope);
                 }
 
                 if (ops.empty()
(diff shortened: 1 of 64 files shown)


include/vast/Translation/CodeGenDeclVisitor.hpp
```diff
@@ -351,14 +344,38 @@ namespace vast::cg {
...
+
+                // Making sure, that if the operation is enclosed in a trailing
+                // scope, then the termiantor is evaluated in this scope (which
+                // will then be spliced by subsequent pass)
+                auto scope = mlir::dyn_cast< hl::ScopeOp >(ops.back());
+
+                while (scope) {
+                    auto parent = scope->getParentRegion();
+                    if (parent->hasOneBlock()
+                        && parent->back().begin() == --parent->back().end())
```
lkorenc, line 362:
```suggestion
                        && parent->back().begin() == std::prev(parent->back().end()))
```

include/vast/Translation/CodeGenDeclVisitor.hpp
```diff
@@ -351,14 +344,38 @@ namespace vast::cg {
...
+                    {
+                        last_op = get_last_op(scope);
+                        set_insertion_point_to_end(&scope.getBody());
+                        if (last_op)
+                            scope = mlir::dyn_cast< hl::ScopeOp >(last_op);
```
lkorenc, line 367: `mlir::dyn_cast` returns an object that evaluates to `false` if it fails, so cannot you simply `scope = dyn_cast< ...` and then the while condition will still work?

Also, this can be a local lambda? To avoid re-assigning variables?

include/vast/Translation/CodeGenDeclVisitor.hpp
```diff
@@ -351,14 +344,38 @@ namespace vast::cg {
...
+                        else
+                            scope = nullptr;
+                    } else {
+                        break;
```
lkorenc, line 371: Early return.

include/vast/Translation/Util.hpp
```diff
@@ -14,46 +14,12 @@ VAST_UNRELAX_WARNINGS
...
-
-            auto next = prev.getNextNode();
-
-            auto &ops = last_block.getOperations();
-            ops.splice(ops.end(), next->getOperations());
-
-            next->erase();
-            scope.erase();
+    template< typename T >
+    void filter(const auto &decls, auto &&yield) {
```
lkorenc, line 18: Return `gap::generator` instead of providing callback?

include/vast/Util/Scopes.hpp
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) 2021-present, Trail of Bits, Inc.
+
+#pragma once
+
+#include <vast/Util/Common.hpp>
+
+namespace vast
+{
+    inline bool is_trailing_scope(operation op) {
```
lkorenc, line 9:
```suggestion
    static inline bool is_trailing_scope(operation op) {
```
And same goes for the one below.

lkorenc (approved):

include/vast/Util/Scopes.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+namespace vast
+{
+    static inline bool is_trailing_scope(operation op) {
+        if (!mlir::isa< hl::ScopeOp >(op))
+            return false;
+        if (auto parent = op->getParentRegion()) {
+            if(parent->hasOneBlock()) {
+                auto &block = parent->back();
+                auto last = --block.end();
```
lkorenc, line 15:
```suggestion
                auto last = std::prev(block.end());
```

lkorenc (approved):

PR #335, opened 2023-07-12
Lowering of sign operations to llvm and improve consistency of typenames

Jezurko:


lib/vast/Conversion/Common/IRsToLLVM.cpp
```diff
@@ -864,32 +864,80 @@ namespace vast::conv::irstollvm
...
+            if (llvm::isa< mlir::FloatType >(arg_type))
+                rewriter.replaceOpWithNewOp< LLVM::FSubOp >(op, zero, arg);
+            else
+                rewriter.replaceOpWithNewOp< LLVM::SubOp >(op, zero, arg);
+
+            return logical_result::success();
+        }
+    };
+
+    struct plus : base_pattern< hl::PlusOp >
```
lkorenc, line 892: Is this not simply a case of `one_to_one` conversion we have a base helper for?

lkorenc (approved):

lib/vast/Conversion/Common/IRsToLLVM.cpp
Jezurko: Nope, the plus op does literally nothing. It's for example:
```
int a = 5;
int b = +a;
```
It is present in the AST, but it isn't present in the LLVM IR afterwards.

lkorenc: `ignore_pattern` base helper then?

Jezurko: Oh… Yeah, that does it.

lkorenc (approved):

PR #337, opened 2023-07-13
Integrate splicing pass into `vast-cc`.

Jezurko:
- Makes `vast-cc` behave as before the patch that added the splicing scope pass
- Adds option to disable the pass in `vast-cc` using `--disable-scope-splicing-pass`

### DIFF
--- a/lib/vast/Dialect/HighLevel/HighLevelOps.cpp
+++ b/lib/vast/Dialect/HighLevel/HighLevelOps.cpp
@@ -268,6 +268,56 @@ namespace vast::hl
         st.addTypes(type);
     }
 
+    mlir::ParseResult IfOp::parse(mlir::OpAsmParser &parser, mlir::OperationState &result) {
+        std::unique_ptr< mlir::Region > condRegion = std::make_unique< mlir::Region >();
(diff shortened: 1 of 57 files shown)


lib/vast/Dialect/HighLevel/HighLevelOps.cpp
```diff
@@ -268,6 +268,58 @@ namespace vast::hl
...
+    mlir::ParseResult IfOp::parse(mlir::OpAsmParser &parser, mlir::OperationState &result) {
+        std::unique_ptr<::mlir::Region> condRegionRegion = std::make_unique<::mlir::Region>();
+        std::unique_ptr<::mlir::Region> thenRegionRegion = std::make_unique<::mlir::Region>();
+        std::unique_ptr<::mlir::Region> elseRegionRegion = std::make_unique<::mlir::Region>();
+
+        if (parser.parseRegion(*condRegionRegion))
+            return ::mlir::failure();
+
+        if (parser.parseKeyword("then"))
+            return mlir::failure();
```
lkorenc, line 280: Inconsistency with above `::mlir`. Why is `::mlir` used here? (Same in the `unique_ptrs`).

lkorenc (approved):

PR #340, opened 2023-07-17
Add option to add new data members to cg stack.

### LINKED ISSUE #339: Add option to add new data members to cg stack.
Currently, the way whole `cg::` stack is implemented, we cannot easily extend it with new data members (we can easily only extend the API).

Sadly, C++ is not doing us a service here with all the templates needed for the CRTP, but basically what needs to be done is to add an option to add type of `_cgctx` as a template. This will sadly bubble up into few classes, so it is not an exactly easy change.

surovic:
Resolves #339

### DIFF
--- a/include/vast/CodeGen/Generator.hpp
+++ b/include/vast/CodeGen/Generator.hpp
@@ -47,9 +47,9 @@ namespace vast::cg {
         void dump_module() { codegen->dump_module(); }
 
     protected:
-        std::unique_ptr< codegen_driver > codegen = nullptr;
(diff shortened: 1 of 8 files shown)


.clang-format
```diff
@@ -0,0 +1 @@
+external/gap/.clang-format
```
lkorenc, line 1: This looks like an accident? At least based on commit messages.

lkorenc (changes requested):

surovic:
Fixes #339

lkorenc (approved):

PR #348, opened 2023-07-25
Initial prototype of the `tower` component.

surovic:
The `tower` component provides an API for applying passes while maintaining a history with provenance information for every operation. This is done by propagating `mlir::Operation *` of the old operation to the new operation via `mlir::FusedLoc` and `mlir::OpaqueLoc`.

### DIFF
--- a/include/vast/Tower/Tower.hpp
+++ b/include/vast/Tower/Tower.hpp
@@ -10,12 +10,6 @@ VAST_UNRELAX_WARNINGS
 
 namespace vast::tw {
 
-    struct handle_t
-    {
-        std::size_t id;
-        vast_module mod;
-    };
-
     struct default_loc_rewriter_t
     {
         static auto insert(mlir::Operation *op) -> void;
@@ -29,6 +23,12 @@ namespace vast::tw {
     struct tower
     {
         using loc_rewriter = loc_rewriter_t;
+
+        struct handle_t
+        {
+            std::size_t id;
+            vast_module mod;
+        };
 
         static auto get(mcontext_t &ctx, owning_module_ref mod)
             -> std::tuple< tower, handle_t > {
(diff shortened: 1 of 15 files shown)


include/vast/Tower/Tower.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+            _modules.emplace_back(mlir::cast< vast_module >(handle.mod->clone()));
+
+            auto id  = _modules.size() - 1;
+            auto mod = _modules.back().get();
+
+            mlir::PassManager pm(&_ctx);
+            pm.addPass(std::move(pass));
+
+            VAST_CHECK(mlir::succeeded(pm.run(mod)), "Some pass in apply() failed");
```
xlauko, line 46: This is not required to be in-release macro.

tools/vast-repl/command.cpp
```diff
@@ -108,11 +109,26 @@ namespace vast::repl::cmd {
...
+                get(state);
+                break;
         }
-    };
+    }
+
+    //
+    // apply command
+    //
+    void apply::run(state_t &state) const {
```
xlauko, line 126: this is not using pass name param

tools/vast-repl/command.cpp
```diff
@@ -108,11 +109,26 @@ namespace vast::repl::cmd {
...
+        check_and_emit_module(state);
+
+        auto [tm, th] = tower::default_manager_t::get(state.ctx, std::move(state.mod));
+
+        llvm::outs() << tm.apply(th, createHLToLLCFPass()).mod << "\n";
```
xlauko, line 131: This replicates functionality of `show module`.

If this is supposed to create new layer of tower lets call it `raise` and we can keep tower in the state.

In fact my idea was that `vast_module` will eventually be replaced with `tower`.
It  shloud be irrelevant whether `tower` has single layer or more.

xlauko (approved):

Issue #347, opened 2023-07-24
Incorrect lowering the function type of copy constructor leads to mlir validation failure


kumarak:
The function type of default copy constructor is getting lowered incorrectly causing the MLIR validation failure due to type conversion issue. Attached generated mlir and source file for reproducing the issue:

error:
```
input.cc:2:9: error: 'hl.ref' op inferred type(s) '!hl.lvalue<!hl.lvalue<!hl.reference<!hl.elaborated<!hl.record<"Student">,  const >>>>' are incompatible with return type(s) of operation '!hl.lvalue<!hl.elaborated<!hl.record<"Student">,  const >>'
```

mlir module dump:
```
"builtin.module"() ({
  "hl.func"() ({
  }) {function_type = (!hl.lvalue<!hl.ptr<!hl.void>>, !hl.lvalue<!hl.ptr<!hl.void< const >>>, !hl.lvalue<!hl.long< unsigned >>) -> !hl.ptr<!hl.void>, linkage = 0 : i32, sym_name = "__builtin_memcpy", sym_visibility = "private"} : () -> ()
  "hl.func"() ({
  ^bb0(%arg0: !hl.lvalue<!hl.lvalue<!hl.reference<!hl.elaborated<!hl.record<"Student">,  const >>>>):
    "hl.scope"() ({
      %0 = "hl.this"() : () -> !hl.ptr<!hl.record<"Student">>
      %1 = "hl.member"(%0) {name = "registry"} : (!hl.ptr<!hl.record<"Student">>) -> !hl.lvalue<!hl.int>
      %2 = "hl.ref"(%arg0) : (!hl.lvalue<!hl.lvalue<!hl.reference<!hl.elaborated<!hl.record<"Student">,  const >>>>) -> !hl.lvalue<!hl.elaborated<!hl.record<"Student">,  const >>
      %3 = "hl.member"(%2) {name = "registry"} : (!hl.lvalue<!hl.elaborated<!hl.record<"Student">,  const >>) -> !hl.lvalue<!hl.int< const >>
      %4 = "hl.implicit_cast"(%3) {kind = 4 : i64} : (!hl.lvalue<!hl.int< const >>) -> !hl.int
      %5 = "hl.assign"(%4, %1) : (!hl.int, !hl.lvalue<!hl.int>) -> !hl.int
      %6 = "hl.this"() : () -> !hl.ptr<!hl.record<"Student">>
[...]

PR #349, opened 2023-07-25
Add unsupported operation visitor as fallback for codegen

### LINKED ISSUE #327: Implement unsupported dialect
Since VAST cannot model every AST (and it will take a while since it will be able) provide a default unsupported dialect.

If we want to be fancy we can provide more unsupported operations, but for now let's begin with a generic operation in ~following shape:

```
%A = vast.unsupported< "some string data to provide more visibility into what failed" >
{
   properly lowered child nodes if any present
}
```

kumarak:
Resolves #327

- Implement unsupported operation dialects
- Add Unsupported Operation visitor as another fallback layer in codegen
- Add support for lowering the unsupported AST node using unsupported op dialects


include/vast/Translation/CodeGenContext.hpp
```diff
@@ -161,6 +161,9 @@ namespace vast::cg
                 if (llvm::isa< clang::FunctionDecl >(dctx))
                     continue;
 
+                if (llvm::isa< clang::LinkageSpecDecl >(dctx))
```
lkorenc, line 164: How is this related to the unsupported dialect? I mean having some extra small-ish change(s) thrown in a bigger PR is fine, but the commit should be self-contained and ideally mention it in the PR description.

include/vast/Dialect/HighLevel/HighLevelUnsupported.td
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) 2022-present, Trail of Bits, Inc.
+
+#ifndef VAST_DIALECT_HIGHLEVEL_IR_HIGHLEVELUNSUPPORTED
+#define VAST_DIALECT_HIGHLEVEL_IR_HIGHLEVELUNSUPPORTED
+
+def HighLevel_UnsupportedDecl
```
lkorenc, line 6: We want to have a separate dialect for unsupported operations.

include/vast/Translation/CodeGen.hpp
```diff
@@ -579,7 +579,7 @@ namespace vast::cg
         }
 
         operation visit_var_decl(const clang::VarDecl *decl) {
-            return _visitor->VisitVarDecl(decl);
+            return _visitor->Visit(decl);
```
lkorenc, line 582: Separate commit.

include/vast/Translation/CodeGenFallBackVisitor.hpp
```diff
@@ -31,21 +32,54 @@ namespace vast::cg
...
         using DeclVisitor = DefaultFallBackDeclVisitorMixin< Derived >;
         using StmtVisitor = DefaultFallBackStmtVisitorMixin< Derived >;
         using TypeVisitor = DefaultFallBackTypeVisitorMixin< Derived >;
 
-        using DeclVisitor::Visit;
-        using StmtVisitor::Visit;
-        using TypeVisitor::Visit;
+        using UnsupportedVisitor = FallBackMixin< Derived >;
+
+        Operation* Visit(const clang::Stmt *stmt) {
```
lkorenc, line 50: I am really confused why we are doing dispatch here again, when a few lines below it is done for us? `CodeGenFallBackVisitorMixin`.

lkorenc: Shouldn't the only change to this class be change of those type alias of partial Visitors to the `Unsupported` versions?

include/vast/Translation/CodeGenFallBackVisitor.hpp
```diff
@@ -31,21 +32,54 @@ namespace vast::cg
...
+        Type Visit(clang::QualType type) {
+            if (auto result = UnsupportedVisitor::Visit(type))
+                return result;
+
+            return TypeVisitor::Visit(type);
+        }
     };
 
+    template< typename Derived >
+    using DefaultFallbackVisitorConfig = DefaultFallBackVisitorMixin< Derived,
```
lkorenc, line 76: Keep the default for now as it was, instead provide extra helpers for unsupported one, so you can easily plug it into the whole stack (I think not trying to use it as a default for now will help flesh out the implementation).

include/vast/Translation/CodeGenUnsupportedVisitor.hpp
```diff
@@ -0,0 +1,527 @@
...
+#include "vast/Translation/CodeGen.hpp"
+#include "vast/Translation/CodeGenContext.hpp"
+#include "vast/Translation/CodeGenVisitor.hpp"
+#include "vast/Translation/CodeGenBuilder.hpp"
+#include "vast/Translation/CodeGenFallBackVisitor.hpp"
+
+namespace vast::cg
+{
+
+#define VISITOR_DECL(decl_type) \
```
lkorenc: Unscoped name that is not properly unique. Also undefine at the end.

include/vast/Translation/CodeGenUnsupportedVisitor.hpp
```diff
@@ -0,0 +1,527 @@
...
+
+		using LensType::derived;
+		using LensType::context;
+		using LensType::mcontext;
+		using LensType::acontext;
+
+		using LensType::meta_location;
+		using LensType::visit;
+
+		template< typename Op, typename... Args >
```
lkorenc: I am pretty sure all of these helpers are copied from other parts of `cg::` (or at least they look similar-ish enough to be considered a good candidates for shared utilities) - cannot they be re-used?

include/vast/Translation/CodeGenUnsupportedVisitor.hpp
```diff
@@ -0,0 +1,527 @@
...
+	{
+	   	using DeclVisitor = UnsupportedDeclVisitor< Derived >;
+	    using StmtVisitor = UnsupportedStmtVisitor< Derived >;
+	    using TypeVisitor = UnsupportedTypeVisitor< Derived >;
+
+	    using DeclVisitor::Visit;
+	    using StmtVisitor::Visit;
+	    using TypeVisitor::Visit;
+
+	    operation Visit(const clang::Stmt *stmt) {
```
lkorenc: Why are these methods here? Except for the `Type` overload, which has custom logic.

include/vast/Translation/CodeGenFallBackVisitor.hpp
```diff
@@ -31,21 +32,54 @@ namespace vast::cg
         }
     };
 
-    template< typename Derived >
+    template< typename Derived
+        , template< typename > typename FallBackMixin = UnsupportedFallBackVisitor
```
lkorenc, line 36: `Unsupported` should not be anywhere in this class -> it should be plugged into the "dispatcher" class below as a template argument only (no extra type aliases or anything should be added to it).

lkorenc (changes requested):

include/vast/Translation/CodeGenContext.hpp
kumarak: If you are walking the decl context parent, you may find `LinkageSpecDecl` as one of the parent and not a `NamedDecl`. That will lead to reaching the unreachable. 

I will move it in the separate commit under this PR to avoid mixup with other changes.

include/vast/Dialect/HighLevel/HighLevelUnsupported.td
kumarak: @lkorenc, Should the unsupported operation dialect named differently. I used `HighLevel_` as prefix because it is high level dialect. The dialects are of three types. The `UnsupportedDecl` is used only for lowering the Decl node. `UnsupportedExpr` and `UnsupportedOp` are used for lowering stmt node.

Could you explain what do you mean by saying "want to have a separate dialect for unsupported operations"?

lkorenc: Yeah I do not want it to be part of high level dialect. I would like a separate dialect - unsupported - same way as we have `ll` and `core` etc. alongside the `hl`. And this unsupported dialect will have those 3 operations.

Or do you think there is a reason this cannot be achieved/is not a good idea?

include/vast/Translation/CodeGenUnsupportedVisitor.hpp
kumarak: undef the macro at the end to avoid conflict with other macros with same name.

include/vast/Dialect/HighLevel/HighLevelUnsupported.td
kumarak: @lkorenc, Thanks for explaining it. We can have `unsupported` as a different set of dialect similar to `core`, or `low-level`. The one concern I have is the coverage scope for these dialects are limited. In that case is it a good idea to have them as separate dialect set? I am not sure.

I will have these changes in a separate commit and we can review how much complexity does it add, if we introduce a new set of dialect for unsupported. We can then decide if we should go ahead with that or not.

include/vast/Translation/CodeGenUnsupportedVisitor.hpp
kumarak: Most of these utility functions are custom and written similar to one in `cg::Builder`. These utility function buildup region and generate operation specific to the information we need to present with the unsupported operations. Moving them to the common place in builder will not be a good idea. 

I may refactor them a few and see if they can use already existing utility methods from `cg::Builder`.

include/vast/Translation/CodeGenUnsupportedVisitor.hpp
kumarak: Removed this additional indirection!

include/vast/Dialect/Unsupported/Unsupported.td
```diff
@@ -0,0 +1,39 @@
...
+include "vast/Interfaces/SymbolInterface.td"
+
+def Unsupported_Dialect : Dialect {
+    let name = "unsupported";
+    let cppNamespace = "::vast::us";
+
+    let summary = "A vast unsupported dialect.";
+
+    let description = [{
+        This dialect intends capture user metadata
```
xlauko, line 19: copy-paste from meta dialect

include/vast/Dialect/Unsupported/UnsupportedTypes.td
```diff
@@ -0,0 +1,18 @@
...
+#define VAST_DIALECT_IR_UNSUPPORTED_TYPES
+
+def UnsupportedType
+    : Unsupported_Type< "Unsupported" > {
+  let mnemonic = "type";
+  let parameters = (ins "Type":$elementType);
+  let assemblyFormat = "`<` $elementType `>`";
+}
+
+def UnsupportedVoidType : Unsupported_Type< "UnsupportedVoid" > {
```
xlauko, line 13: when do you need such type?

include/vast/Translation/CodeGenFallBackVisitor.hpp
```diff
@@ -56,13 +58,16 @@ namespace vast::cg
     template< typename Derived
         , template< typename > typename DefaultMixin
         , template< typename > typename FallBackMixin = DefaultFallBackVisitorMixin
+        , template< typename > typename UnsupporteFallbackdMixin = UnsupportedFallBackVisitor
```
xlauko, line 61: typo Fallbackd -> Fallback

include/vast/Translation/CodeGenFallBackVisitor.hpp
```diff
@@ -56,13 +58,16 @@ namespace vast::cg
...
     >
     struct CodeGenFallBackVisitorMixin
         : DefaultMixin< Derived >
         , FallBackMixin< Derived >
+        , UnsupporteFallbackdMixin< Derived >
```
xlauko, line 66: This should be parameter not part of this class, ie.e you pass `UnsupporteFallbackdMixin` as `FallBackMixin`.

I can fix this and integrate to codegen with option if you like?

kumarak: @xlauko, Do you mean creating a config that takes `UnsupportedFallBackVisitor` as one of the parameter. Something like defining the config as:
```
    template< typename Derived >
    using DefaultCodeGenVisitorConfig = CodeGenFallBackVisitorMixin< Derived,
        DefaultCodeGenVisitorMixin,
        UnsupportedFallBackVisitor
    >;
```

One of the reason I didn't add it to `DefaultCodeGenVisitorConfig` because I want both unsupported visitor and default fallback visitor. Previous I updated `DefaultFallBackVisitorMixin` and passed a config as parameter that will multiplex between Unsupported Visitor and Default fallback visitor. It was similar to something like
```
template< typename Derived >
    using DefaultFallbackVisitorConfig = DefaultFallBackVisitorMixin< Derived, UnsupportedFallBackVisitor>;

// pass config to DefaultCodeGenVisitorConfig

    template< typename Derived >
    using DefaultCodeGenVisitorConfig = CodeGenFallBackVisitorMixin< Derived,
        DefaultCodeGenVisitorMixin,
        DefaultFallbackVisitorConfig
    >;
```

This will allow choosing between the default fallback and unsupported visitor in class `DefaultFallBackVisitorMixin` but that mean adding another redirection there.

Can you write a pseudo code of how you like it to be integrated? I will try getting that working else will ask you to update the update it with the options.

xlauko: Thats exactly how it was designed to be used. You don't need to choose anything in `DefaultFallBackVisitorMixin`.
FallBack mixin takes care of trying all visitors in order.

Under the hood it will generate the same code as you have now, which essentially cascades visits `default -> unsupported -> default error`.

include/vast/Dialect/HighLevel/HighLevelUnsupported.td
xlauko: I think we can remove this now. Unsupported dialect is the way to go, because we can have eventually something that does not use `hl` dialect but still leverage `unsupported` stuff.

include/vast/Dialect/Unsupported/UnsupportedTypes.td
xlauko: I think it can be merged to single `UnsupportedType` where you make `elementType` optional:

```
  let assemblyFormat = "`<` ( $elementType^ )? `>`";
```

or at least change the name as `UnsupportedVoidType` is a bit confusing.
Use something like `UnsupportedOpaqueType`

include/vast/Translation/CodeGenFallBackVisitor.hpp
kumarak: @xlauko, I added `FallbackVisitorConfig` to chain the unsupported op fallback. I think this is how it is intended to be chained with `DefaultFallBackVisitorMixin`. Let me know if it looks good.

include/vast/Dialect/Unsupported/UnsupportedTypes.td
kumarak: I removed `UnsupportedVoidType` and used the `OptionalParameter` for `elementType`. We can use it for generating types in opaque format as well where we don't visit the underlying types.

xlauko:
Closes #366 #365

lkorenc (approved):

lkorenc:
Some commits are a mess, I would be glad if they were cleaned up.

kumarak:
@xlauko, Refactoring looks good but I do see it hits llvm unreachable during binding the callback values with the operations. I am seeing with most of my unit test cases whenever it reaches Unsupported visitor.

```
UNREACHABLE executed at /Users/akkumar/Documents/workspace/cxx-common/vcpkg/buildtrees/llvm-16/src/org-16.0.5-90a3d25b70.clean/clang/lib/AST/DeclBase.cpp:946!
```

xlauko:
@kumarak can you provide your tests somewhere please.

kumarak:
Unit test looks good to me now. I see one failure but that is related to lowering c++ default copy constructor #347 and not unsupported visitor.

lkorenc (approved):

xlauko (approved):

xlauko (approved):

Issue #338, opened 2023-07-14
Implement error reporting via exceptions.


lkorenc:
Currently we are using `VAST_[ASSERT, CHECK, UNREACHABLE]` - all of which trigger a hard failure if fired. This is however not the only mode of functionality we want to support, as we do not want to kill programs that are using our libs just because we encountered and error.

Implement a function that instead of killing the program throws vast exception. Then once this is done provide a compile-time option that allows users to choose between hard error or exception throwing as their mechanism of choice.

PR #350, opened 2023-07-31
Support error reporting via exceptions

kumarak:
Fixes issue #338

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -165,11 +165,11 @@ if (NOT LLVM_ENABLE_RTTI)
 endif()
 
 # enable exception on failure
-option(ENABLE_VAST_EXCEPTION "Enable exception throw on vast failure" OFF)
-if(ENABLE_VAST_EXCEPTION)
+option(VAST_ENABLE_EXCEPTIONS "Enable exception throw on vast failure" OFF)
+if (VAST_ENABLE_EXCEPTIONS)
   target_compile_definitions(vast_settings
     INTERFACE
-      -DENABLE_VAST_EXCEPTION
+      -DVAST_ENABLE_EXCEPTIONS
   )
 
   target_compile_options(vast_settings
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -164,6 +164,20 @@ if (NOT LLVM_ENABLE_RTTI)
   )
 endif()
 
+# enable exception on failure
+option(ENABLE_VAST_EXCEPTION "Enable exception throw on vast failure" OFF)
+if(ENABLE_VAST_EXCEPTION)
+  target_compile_definitions(vast_settings
+    INTERFACE
+      -DENABLE_VAST_EXCEPTION
+  )
+
+  target_compile_options(vast_settings
+    INTERFACE
+      -fexceptions
+  )
+endif()
+
 # sanitizer options if supported by compiler
 include(cmake/sanitizers.cmake)
 enable_sanitizers(vast_settings)
--- a/include/vast/Util/Warnings.hpp
+++ b/include/vast/Util/Warnings.hpp
@@ -47,7 +47,7 @@ VAST_RELAX_WARNINGS
 #include <llvm/Support/Debug.h>
 VAST_UNRELAX_WARNINGS
 
-#ifdef ENABLE_VAST_EXCEPTION
+#ifdef VAST_ENABLE_EXCEPTIONS
 #include <stdexcept>
 #include <sstream>
 
--- a/include/vast/Util/Warnings.hpp
+++ b/include/vast/Util/Warnings.hpp
@@ -54,23 +54,22 @@ VAST_UNRELAX_WARNINGS
 namespace vast
 {
     struct error_stream {
-      private:
-        std::stringstream buff;
+        explicit error_stream() : ss(buff) {}
 
-       public:
-        error_stream(std::stringstream ss = std::stringstream())
-              : buff(std::move(ss)) {}
         ~error_stream() noexcept(false) {
-            throw std::runtime_error(buff.str());
+            ss.flush();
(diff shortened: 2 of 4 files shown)


include/vast/Util/Warnings.hpp
```diff
@@ -47,16 +47,54 @@ VAST_RELAX_WARNINGS
...
+#define vast_debug llvm::dbgs
+
 namespace vast
 {
+#ifdef ENABLE_VAST_EXCEPTION
+    template <typename T = std::runtime_error>
+    struct error {
+      private:
+        std::string msg;
+        bool triggered = false;
```
lkorenc, line 69: I am confused about the name, because if we look in the `dtor`, it only throws `if triggered is true`. So either rename it or fix the values.

include/vast/Util/Warnings.hpp
```diff
@@ -47,16 +47,54 @@ VAST_RELAX_WARNINGS
...
+
+       public:
+        error(bool cond = true, std::string msg_ = std::string())
```
lkorenc, line 72: Why is there and option to set `triggered`? Is there a use-case for passing `false`?

include/vast/Util/Warnings.hpp
```diff
@@ -47,16 +47,54 @@ VAST_RELAX_WARNINGS
...
+              : msg(std::move(msg_)), triggered(cond) {}
```
lkorenc, line 73: Argument can be just named `msg` (while I know some people like the codestyle of adding the `_` to the ctor VAST does not do it).

include/vast/Util/Warnings.hpp
```diff
@@ -47,16 +47,54 @@ VAST_RELAX_WARNINGS
...
+        ~error() noexcept(false) {
+            if (triggered) {
+              throw T(msg);
+            }
+        }
+
+        template <typename V>
+        error<T> operator<<(V&& s) {
+            msg += std::forward<V>(s);
+            return error<T>(triggered, msg);
```
lkorenc, line 83:
```suggestion
            return error<T>(triggered, std::move(msg));
```

include/vast/Util/Warnings.hpp
```diff
@@ -47,16 +47,54 @@ VAST_RELAX_WARNINGS
...
+    struct error {
+      private:
+        std::string msg;
+        bool triggered = false;
+
+       public:
+        error(bool cond = true, std::string msg_ = std::string())
+              : msg(std::move(msg_)), triggered(cond) {}
+        ~error() noexcept(false) {
+            if (triggered) {
```
lkorenc, line 75: You are never setting this to `false` (and in `operator<<` there is a new object being made) so in move ctor and move assign we should set it to `false`, otherwise that's a bug once multiple times `<<` is invoked.

include/vast/Util/Warnings.hpp
lkorenc: Why not just return reference to `this`, make no copies/moves (as in deleted ctors/`=`) and then simply always throw in the dtor? Should streamline this.

lkorenc (dismissed):

CMakeLists.txt
```diff
@@ -176,6 +185,7 @@ if (ENABLE_PDLL_CONVERSIONS)
   message(STATUS "ENABLE_PDLL_CONVERSIONS")
   target_compile_definitions(vast_settings
     INTERFACE
+      -fexceptions
```
xlauko, line 188: add this only in case of `ENABLE_VAST_EXCEPTION` is `ON`

move it to previous `target_compile_definitions`

include/vast/Util/Warnings.hpp
```diff
@@ -47,16 +47,54 @@ VAST_RELAX_WARNINGS
 #include <llvm/Support/Debug.h>
 VAST_UNRELAX_WARNINGS
 
+#include <stdexcept>
```
xlauko, line 50: move under `ENABLE_VAST_EXCEPTION`

include/vast/Util/Warnings.hpp
```diff
@@ -47,16 +47,54 @@ VAST_RELAX_WARNINGS
...
+  #define vast_error llvm::dbgs
+#endif
+
+#define vast_debug llvm::dbgs
+
 namespace vast
 {
+#ifdef ENABLE_VAST_EXCEPTION
+    template <typename T = std::runtime_error>
+    struct error {
```
xlauko, line 66: rename to `error_stream`

xlauko: also something like:
```
struct error_stream {
    std::stringstream buff;
    
    virtual ~error_stream() noexcept(false) {
        throw std::runtime_error(buff.str());
    }

    template< typename token_type >
    error_stream& operator<<(token_type &&token) {
        buff << std::forward< token_type >(token);
        return *this;
    }
};
```

should work?

CMakeLists.txt
kumarak: Oops, my mistake. I was intended to add under `ENABLE_VAST_EXCEPTION` but added under different flag. Will fix this.

include/vast/Util/Warnings.hpp
kumarak: Removed `triggered` flag from the `error` class. My initial thought was to enable the exception only if `triggered` is set to true. But probably we don't need to have all these complexity and should have simple implementation of `error_stream`.

include/vast/Util/Warnings.hpp
kumarak: removed from the changes

kumarak:
@lkorenc, @xlauko, I addressed the comments and simplified the implementation of `error_stream` whose only job is to throw the exception in `dtor`. Having template and trigger flag was probably making it look complex.

xlauko (approved):

PR #362, opened 2023-08-03
Add header with some hl utilities

lkorenc:
Introduce a bunch of utilities for working with records. These were before randomly re-implemented in a bunch of conversions. So far I am keeping them header only as I suspect we will need to eventually change some concerte `hl::` types to templates anyway to support C++ as well.

PR also includes refactor of some of our more in-progress passes to use this newer api.

There is one sneaky `ll` commit that adds `Extract` and `Concat` operations which will be needed by abi lowering.

### DIFF
--- a/include/vast/Conversion/Common/Rewriter.hpp
+++ b/include/vast/Conversion/Common/Rewriter.hpp
@@ -10,6 +10,11 @@ VAST_UNRELAX_WARNINGS
 
 #include "vast/Util/Common.hpp"
 
+#include <gap/core/generator.hpp>
+
+#include <ranges>
+#include <vector>
+
 namespace vast::conv
 {
     template< typename impl_t >
@@ -61,6 +66,24 @@ namespace vast::conv
             auto g = guard();
             bld.setInsertionPointToEnd( block );
             return bld.template create< Trg >( std::forward< Args >( args ) ... );
+        }
+
+        template< typename T >
+        void safe_erase( gap::generator< T > &&range )
+        {
+            // It is highly probable there is an ongoing iteration
+            // inside the generator, therefore we first freeze the value.
+            // TODO(c++23): Refactor.
+            std::vector< T > to_erase;
+            std::ranges::move(range.begin(), range.end(), std::back_inserter(to_erase));
+            return erase( to_erase );
+        }
+
+        template< typename O >
+        void erase( const std::vector< O > &to_erase )
+        {
+            for ( auto o : to_erase )
+                bld.eraseOp( o );
         }
     };
 
--- a/include/vast/Conversion/Common/Rewriter.hpp
+++ b/include/vast/Conversion/Common/Rewriter.hpp
@@ -75,7 +75,7 @@ namespace vast::conv
             // inside the generator, therefore we first freeze the value.
             // TODO(c++23): Refactor.
             std::vector< T > to_erase;
-            std::ranges::move(range.begin(), range.end(), std::back_inserter(to_erase));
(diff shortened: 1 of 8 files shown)


lib/vast/Conversion/FromHL/StructsToLLVM.cpp
```diff
@@ -50,66 +53,35 @@ namespace vast
...
-
-                rewriter.create< hl::TypeDefOp >(
-                        op.getLoc(), op.getName(), trg_ty);
-
-                auto type_decls = fetch_decls(op);
-                for (auto x : type_decls)
-                    rewriter.eraseOp(x);
+                auto llvm_struct = convert_struct_type();
+                rewriter.create< hl::TypeDefOp >(op.getLoc(), op.getName(), llvm_struct);
+                // TODO(conv:hl-structs-to-llvm): Investigate if this is still required.
```
xlauko, line 83: add issue

include/vast/Dialect/HighLevel/HighLevelTypes.hpp
```diff
@@ -38,6 +38,9 @@ namespace vast::hl
     mlir::Type strip_elaborated(mlir::Type);
     mlir::Type strip_elaborated(mlir::Value);
 
+    mlir::Type strip_value_category(mlir::Type);
+    mlir::Type strip_value_category(mlir::Value);
+
```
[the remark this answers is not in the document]
lkorenc: Well my idea is that this one will eventually strip all value categories (`lvalue`, `rvalue`, etc). I can add an overload that combines this with `strip_elaborated`. Or did you mean something else?
In general I am not sure yet what will be things that we will need removing often, so I started with basic building blocks.

include/vast/Dialect/HighLevel/HighLevelUtils.hpp
```diff
@@ -0,0 +1,117 @@
...
+        }
+    }
+
+    static inline type_generator field_types(hl::StructDeclOp op)
+    {
+        for (auto def : field_defs(op))
+            co_yield def.getType();
+    }
+
+    static inline auto definition_of(mlir::Type t, vast_module mod)
```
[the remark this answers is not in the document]
lkorenc: `TODO`. I will open an issue. This is a first step in generalising these helpers (before they were at about 3 random places). The scoping issue will be applicable to all functions in this header.

include/vast/Conversion/Common/Rewriter.hpp
```diff
@@ -62,6 +67,24 @@ namespace vast::conv
...
         }
+
+        template< typename T >
+        void safe_erase( gap::generator< T > &&range )
+        {
+            // It is highly probable there is an ongoing iteration
+            // inside the generator, therefore we first freeze the value.
+            // TODO(c++23): Refactor.
+            std::vector< T > to_erase;
+            std::ranges::move(range.begin(), range.end(), std::back_inserter(to_erase));
```
xlauko, line 78: you know the purpose of ranges is not to write `begin` and `end` :P 

`std::ranges::move(range, std::back_inserter(to_erase));` should do

Also I suggest to rename `range` to `ops`.

xlauko (approved):

include/vast/Conversion/Common/Rewriter.hpp
lkorenc: Tell that to few old days younger version of yourself that forgot to even include the `ranges::` subpart of the namespace :P Will fix it, did not realise that I am stepping over to ranges proper instead of old-style api.

PR #363, opened 2023-08-07
Add support for custom codegen contexts in visitor stack.

### LINKED ISSUE #361: Custom codegen context cannot be easily used in codegen stack.
First of all, `context()` method has a baked-in type, which is incorrect.

Secondly, there is no way to specify the type of the context for the class that's get instantiated by the user. Consider I want to use my custom context with `DefaultCodegen` - this is currently not easily possible and we want it to be if we can.

surovic:
Resolves #361

### DIFF
--- a/include/vast/Translation/CodeGenVisitor.hpp
+++ b/include/vast/Translation/CodeGenVisitor.hpp
@@ -2,6 +2,7 @@
 
 #pragma once
 
+#include "vast/Translation/CodeGenScope.hpp"
 #include "vast/Util/Warnings.hpp"
 
 #include "vast/Translation/CodeGenDeclVisitor.hpp"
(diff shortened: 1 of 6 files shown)


include/vast/Translation/CodeGenVisitor.hpp
```diff
@@ -40,13 +41,13 @@ namespace vast::cg
     //
     // `MetaGenerator` takes care of attaching location metadata to generated mlir primitives.
     //
-    template< template< typename > class CodeGenVisitorMixin, MetaGeneratorLike MetaGenerator >
+    template< typename CodeGenContext, template< typename > class CodeGenVisitorMixin, MetaGeneratorLike MetaGenerator >
```
xlauko, line 44: Umaybe just use `Context` or `CGContext` not to shadow the class `CodeGenContext`.

lib/vast/Translation/FromSource.cpp
```diff
@@ -60,7 +60,7 @@ namespace vast::hl
         if (id_meta_flag) {
             cg::CodeGenWithMetaIDs(cgctx).emit_module(ast.get());
         } else {
-            cg::DefaultCodeGen(cgctx).emit_module(ast.get());
+            cg::DefaultCodeGen<cg::CodeGenContext>(cgctx).emit_module(ast.get());
```
xlauko, line 63: This should be deducible from constructor parameter. Add deduction guide if necessery.

tools/vast-repl/codegen.cpp
```diff
@@ -30,7 +30,7 @@ namespace vast::repl::codegen {
         auto unit = codegen::ast_from_source(source);
         auto &actx = unit->getASTContext();
         vast::cg::CodeGenContext cgctx(*mctx, actx);
-        vast::cg::DefaultCodeGen codegen(cgctx);
+        vast::cg::DefaultCodeGen<vast::cg::CodeGenContext> codegen(cgctx);
```
xlauko, line 33: Same here: "This should be deducible from constructor parameter. Add deduction guide if necessery."

include/vast/Translation/CodeGenVisitor.hpp
xlauko: Similarly in other templates.

xlauko (approved):

PR #368, opened 2023-08-11
Fix VarDecl where the variable is used on the right side and change hl regions to graph regions

Jezurko:
e.g.: `int *x = malloc(sizeof(*x));`

These changes also require us to make any region in `hl` that may contain `VarDecl` a graph region (TODO).

### DIFF
--- a/include/vast/CodeGen/CodeGenDeclVisitor.hpp
+++ b/include/vast/CodeGen/CodeGenDeclVisitor.hpp
@@ -482,6 +482,9 @@ namespace vast::cg {
                     .bind(meta_location(decl))                                  // location
                     .bind(visit_as_lvalue_type(type))                           // type
                     .bind(context().decl_name(decl->getUnderlyingDecl()))       // name
+                    // The initializer region is filled later as it might
+                    // have references to the VarDecl we are currently
+                    // visiting - int *x = malloc(sizeof(*x))
                     .bind_region_if(has_init, [](auto, auto){})                 // initializer
                     .bind_region_if(has_allocator, std::move(array_allocator))  // array allocator
                     .freeze();
--- a/include/vast/CodeGen/CodeGenDeclVisitor.hpp
+++ b/include/vast/CodeGen/CodeGenDeclVisitor.hpp
@@ -466,12 +466,11 @@ namespace vast::cg {
         }
 
         operation VisitVarDecl(const clang::VarDecl *decl) {
-            return context().declare(decl, [&] {
+            auto var_decl = context().declare(decl, [&] {
                 auto type = decl->getType();
                 bool has_allocator = type->isVariableArrayType();
                 bool has_init = decl->getInit();
 
-                auto initializer = make_value_builder(decl->getInit());
 
                 auto array_allocator = [decl, this](auto &bld, auto loc) {
                     if (auto type = clang::dyn_cast< clang::VariableArrayType >(decl->getType())) {
@@ -483,7 +482,7 @@ namespace vast::cg {
                     .bind(meta_location(decl))                                  // location
                     .bind(visit_as_lvalue_type(type))                           // type
                     .bind(context().decl_name(decl->getUnderlyingDecl()))       // name
-                    .bind_region_if(has_init, std::move(initializer))           // initializer
+                    .bind_region_if(has_init, [](auto, auto){})                 // initializer
                     .bind_region_if(has_allocator, std::move(array_allocator))  // array allocator
                     .freeze();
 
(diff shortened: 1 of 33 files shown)


include/vast/CodeGen/CodeGenDeclVisitor.hpp
```diff
@@ -483,7 +482,7 @@ namespace vast::cg {
                     .bind(meta_location(decl))                                  // location
                     .bind(visit_as_lvalue_type(type))                           // type
                     .bind(context().decl_name(decl->getUnderlyingDecl()))       // name
-                    .bind_region_if(has_init, std::move(initializer))           // initializer
+                    .bind_region_if(has_init, [](auto, auto){})                 // initializer
```
xlauko, line 485: add comment that region is filled later because of possible self reference

include/vast/Conversion/Passes.td
```diff
@@ -125,6 +125,16 @@ def HLToLLVars : Pass<"vast-hl-to-ll-vars", "mlir::ModuleOp"> {
...
 }
 
+def HLToLLFunc : Pass<"vast-hl-to-ll-func", "mlir::ModuleOp"> {
+  let summary = "Convert hl functions into ll versions.";
+  let description = [{
+    This pass is still a work in progress.
+  }];
+
+  let constructor = "vast::createHLToLLFuncPass()";
+  let dependentDialects = ["mlir::LLVM::LLVMDialect", "vast::ll::LowLevelDialect"];
```
xlauko, line 135: probably also Core dialect?

include/vast/Dialect/HighLevel/HighLevelCF.td
```diff
@@ -79,7 +79,7 @@ def HighLevel_TypeYieldOp : HighLevel_Op< "type.yield", [
   let assemblyFormat = [{ attr-dict $result `:` type($result) }];
 }
 
-def HighLevel_IfOp : ControlFlowOp< "if" >
+def HighLevel_IfOp : ControlFlowOp< "if", [DeclareOpInterfaceMethods<RegionKindInterface>] >
```
xlauko, line 82: Cn not `[DeclareOpInterfaceMethods<RegionKindInterface>]` be part of `ControlFlowOp`?

include/vast/Dialect/HighLevel/HighLevelOps.td
```diff
@@ -29,7 +26,7 @@ def TranslationUnitOp
   let assemblyFormat = [{ $body attr-dict }];
 }
 
-def ScopeOp : HighLevel_Op< "scope", [NoTerminator] >
+def ScopeOp : HighLevel_Op< "scope", [NoTerminator, DeclareOpInterfaceMethods<RegionKindInterface>] >
```
xlauko, line 29: I know this is not part of this PR, but ScopeOp should be part of Core, as it is not actual AST primitive.

include/vast/Dialect/HighLevel/HighLevelOps.td
```diff
@@ -43,41 +40,18 @@ def ScopeOp : HighLevel_Op< "scope", [NoTerminator] >
...
-    OptionalAttr<StrAttr>:$sym_visibility,
-    OptionalAttr<DictArrayAttr>:$arg_attrs,
-    OptionalAttr<DictArrayAttr>:$res_attrs
-  );
-
-  let regions = (region AnyRegion:$body);
+def HighLevel_FuncOp
+  : FuncOp< HighLevel_Dialect
+            , "func"
+            , [DeclareOpInterfaceMethods<RegionKindInterface>]
```
xlauko, line 46: Again Interface can be part of `FuncOp`

include/vast/Dialect/LowLevel/LowLevelOps.td
```diff
@@ -257,4 +260,45 @@ def InlineScope
     let regions = (region AnyRegion:$body);
 }
 
+def LowLevel_FuncOp : FuncOp< LowLevel_Dialect, "func", [] >
+{
+  let skipDefaultBuilders = 1;
+
+  let builders = [OpBuilder< (ins
```
xlauko, line 267: Does function builder differ anyhow between `HL_FUncOp` `LL_FuncOp` and `HLABI_FuncOp` if not thay can be moved to base?

test/vast/Dialect/HighLevel/vars-c.c
```diff
@@ -0,0 +1,49 @@
+// RUN: vast-front -vast-emit-mlir=hl -o - %s
```
xlauko, line 1: add some checks?

include/vast/Conversion/Passes.td
Jezurko: I don't think so actually. Afaik it should contain dialects from which the pass creates new entities. I should probably remove the llvm dialect that I copied form somewhere else though.

include/vast/Dialect/HighLevel/HighLevelOps.td
Jezurko: What do you mean by this? That in codegen we should emit `core.scope` instead of `hl.scope`?

include/vast/Dialect/LowLevel/LowLevelOps.td
Jezurko: `ABIFuncBase` does use different builder from the other two.

include/vast/Dialect/HighLevel/HighLevelOps.td
Jezurko: Okay, I will do it in a separate PR as this is already getting quite big?

xlauko (approved):

PR #371, opened 2023-08-16
Basic ABI to LLVM conversion.

lkorenc:
This PR adds bits and pieces needed to change type of `VAST`-emitted functions to be the same `clang` emits. Therefore, it introduces:
 - [ ] Conversion `abi. -> hl/ll` - only `abi.direct` is currently supported.
 - [ ] Required additions/changes to `irs-to-llvm` to support operations that can be created during first step
 - [ ] Tests that try to compile & link with a `clang` produced object file.

Currently known issues that may end up as part of this PR:
 * Function declarations are not converted.


include/vast/Conversion/ABI/AggregateTypes.hpp
```diff
@@ -0,0 +1,286 @@
...
+            : state(state),
+              mod(mod)
+        {}
+
+        static state_t mk_state(const pattern &parent, op_t abi_op)
+        {
+            return state_t(parent, abi_op);
+        }
+
+        bool needs_nesting(mlir_type type) const
```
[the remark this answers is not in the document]
xlauko: Isn't this in fact obsolete as of `mlir::AttrTypeRewriter`?

include/vast/Dialect/LowLevel/LowLevelOps.td
```diff
@@ -85,27 +85,36 @@ def Extract
     }];
 
     let builders = [
-        OpBuilder<(ins "mlir::Type":$type, "mlir::Type":$attr_type,
+        OpBuilder<(ins "mlir::Type":$type,
                        "mlir::Value":$value, "std::size_t":$from, "std::size_t":$to),
         [{
+            auto attr_type = mlir::IntegerType::get($_builder.getContext(),
+                                                    64, mlir::IntegerType::Unsigned);
```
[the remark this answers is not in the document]
xlauko: Also if, it's always `IntegerAttr` you can update definition of `Extract` so you don't need to cast in the getters later?

include/vast/Util/DataLayout.hpp
```diff
@@ -29,61 +29,97 @@ namespace vast::dl
...
+
+            auto named_attr = dict_attr.getNamed(key);
+            VAST_ASSERT(named_attr);
+
+            auto int_attr = mlir::dyn_cast< mlir::IntegerAttr >(named_attr->getValue());
+            return static_cast< bitwidth_t >(int_attr.getInt());
         }
 
-        static bitwidth_t unwrap_bw(const mlir::Attribute attr)
+        mlir::StringAttr wrap_str(mcontext_t &mctx, llvm::StringRef str) const
```
xlauko, line 72: rename to str_attr

lib/vast/Conversion/ABI/EmitABI.cpp
```diff
@@ -117,9 +117,18 @@ namespace vast
                 return out;
             }
 
-            core::FunctionType abified_type() {
-                // TODO: deal with varaargs
-                return core::FunctionType::get(abified_args(), abified_rets());
+            core::FunctionType abified_type()
+            {
+                auto add_lvalue = [&](const auto &from)
```
xlauko, line 122: this might be nicer with `std::views::transform`, but I am not sure it will cooperate with MLIR Type constructors.

lib/vast/Conversion/ABI/EmitABI.cpp
```diff
@@ -134,21 +143,31 @@ namespace vast
...
+            void zip_ret(const auto &abi, const auto &results, const auto &concretes,
+                         auto &&yield)
             {
-                auto abi_it = a.begin();
-                auto args_it = b.begin();
+                // These should be always synced, but we are defensive as this may
+                // be called with some random module that is not a result of our
+                // classification.
+                VAST_ASSERT(std::ranges::size(abi) == std::ranges::size(results));
+                auto abi_it = abi.begin();
```
xlauko, line 153: what about `llvm::zip`, probably can be applied also in the previous `zip` function

include/vast/ABI/Classify.hpp
```diff
@@ -213,6 +215,8 @@ namespace vast::abi
             VAST_CHECK( type_size != 0, "Unexpected empty field? Type: {0}", t );
 
             auto final_size = std::min< std::size_t >( type_size - ( root_offset * 8 ), 64 );
+            // Creation of `i0` is probably never fine?
+            VAST_ASSERT( final_size != 0 );
```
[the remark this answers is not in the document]
lkorenc: I will open an issue for this and update the `ASSERT` to `CHECK` with proper message.

include/vast/Conversion/ABI/AggregateTypes.hpp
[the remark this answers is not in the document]
lkorenc: Well this is sadly not that simple, since I am pretty sure arrays will be treated same way as any scalar value and we cannot nest here. I will update the docs and merge this with the same functionality below to have it only at one place, but I do not think we can easily now re-use some existing MLIR abstraction.

lib/vast/Conversion/ABI/LowerABI.cpp
```diff
@@ -34,23 +35,632 @@ VAST_UNRELAX_WARNINGS
...
+                                           conversion_rewriter &rewriter) const override
+            {
+                return base::rewrite({ op, ops, rewriter });
+            }
+
+            // TODO(conv:abi): Copy & paste of prologue
+            values match_on(abi::DirectOp direct, state_capture &state) const override
+            {
+                auto dtor = deconstructs_types(state, *this);
+                for (auto v : dtor.handle(direct))
```
xlauko, line 422: why don't you directly yield the `dtor.handle(direct)`?

lib/vast/Conversion/ABI/LowerABI.cpp
```diff
@@ -34,23 +35,632 @@ VAST_UNRELAX_WARNINGS
...
+                                           conversion_rewriter &rewriter) const override
+            {
+                return base::rewrite({ op, ops, rewriter });
+            }
+
+            // TODO(conv:abi): Copy & paste of prologue
+            values match_on(abi::DirectOp direct, state_capture &state) const override
+            {
+                auto ctor = reconstructs_types(state, *this);
+                for (auto v : ctor.handle(direct))
```
xlauko, line 448: same here, you can return coroutine directly?

lib/vast/Conversion/ABI/LowerABI.cpp
```diff
@@ -34,23 +35,632 @@ VAST_UNRELAX_WARNINGS
...
+                                           conversion_rewriter &rewriter) const override
+            {
+                return base::rewrite({ op, ops, rewriter });
+            }
+
+            // TODO(conv:abi): Copy & paste of prologue
+            values match_on(abi::DirectOp direct, state_capture &state) const override
+            {
+                auto dtor = deconstructs_types(state, *this);
+                for (auto v : dtor.handle(direct))
```
xlauko, line 473: same here, you can return coroutine directly?

lib/vast/Conversion/ABI/LowerABI.cpp
```diff
@@ -34,23 +35,632 @@ VAST_UNRELAX_WARNINGS
...
+                                           conversion_rewriter &rewriter) const override
+            {
+                return base::rewrite({ op, ops, rewriter });
+            }
+
+            // TODO(conv:abi): Copy & paste of prologue
+            values match_on(abi::DirectOp direct, state_capture &state) const override
+            {
+                auto ctor = reconstructs_types(state, *this);
+                for (auto v : ctor.handle(direct))
```
xlauko, line 497: same here, you can return coroutine directly?

lib/vast/Conversion/ABI/LowerABI.cpp
```diff
@@ -34,23 +35,632 @@ VAST_UNRELAX_WARNINGS
...
+
+                auto name = op.getName();
+                if (!name.consume_front("vast.abi"))
+                    return mlir::failure();
+
+                auto fn = rewriter.create< hl::FuncOp >(
+                        op.getLoc(),
+                        name,
+                        op.getFunctionType(),
+                        core::GlobalLinkageKind::InternalLinkage,
```
xlauko, line 575: you change linkage here, that differ from to original `op`

lib/vast/Conversion/ABI/LowerABI.cpp
```diff
@@ -34,23 +35,632 @@ VAST_UNRELAX_WARNINGS
...
+                        other_attrs,
+                        arg_attrs
+                );
+
+                fn.setVisibility(mlir::SymbolTable::Visibility::Private);
```
xlauko, line 580: you change visibility here, that differ from to original `op`

lib/vast/Conversion/Common/IRsToLLVM.cpp
```diff
@@ -46,6 +46,176 @@ namespace vast::conv::irstollvm
...
+            for (std::size_t i = 1; i < ops.getOperands().size(); ++i)
+            {
+                auto full = resize(ops.getOperands()[i]);
+                auto shifted = rewriter.create< mlir::LLVM::ShlOp >(
+                        loc,
+                        full,
+                        mk_index(loc, start, rewriter));
+                head = rewriter.create< mlir::LLVM::OrOp >(
+                        loc,
+                        head->getResults()[0],
```
xlauko, line 202: use `getResult(0)`

lib/vast/Conversion/Common/IRsToLLVM.cpp
```diff
@@ -46,6 +46,176 @@ namespace vast::conv::irstollvm
...
+                        shifted->getResults()[0]);
+
+                start += bw(ops.getOperands()[i].getDefiningOp());
```
xlauko, line 205: use `getOperand(i)`

lib/vast/Conversion/Common/IRsToLLVM.cpp
```diff
@@ -159,41 +326,65 @@ namespace vast::conv::irstollvm
...
+        auto erase(auto, auto &) const {}
+        auto erase(hl::InitListExpr op, auto &rewriter) const
         {
-            auto index_type = tc.convert_type_to_type(rewriter.getIndexType());
-            VAST_PATTERN_CHECK(index_type, "Was not able to convert index type");
+            rewriter.eraseOp(op);
+        }
 
-            for (auto element : ops.getElements())
+        static bool is_scalar(mlir_type t)
```
xlauko, line 338: misleading name ... should be something like `points_to_scalar`?

lib/vast/Conversion/Common/IRsToLLVM.cpp
```diff
@@ -159,41 +326,65 @@ namespace vast::conv::irstollvm
...
+                    element.getLoc(),
+                    element,
+                    ptr);
+        }
+
+        void handle_init_list(auto init_list, auto ptr, auto &rewriter) const
+        {
+            std::size_t i = 0;
+
+            for (auto element : init_list.getElements())
```
xlauko, line 366: use `llvm::enumerate` instead of random local index variable

lib/vast/Target/LLVMIR/Convert.cpp
```diff
@@ -30,6 +30,52 @@ VAST_UNRELAX_WARNINGS
...
+    namespace
+    {
+        // TODO(target): Unify with tower and opt.
+        void populate_pm(mlir::PassManager &pm, pipeline p)
+        {
+            switch (p)
+            {
+                case pipeline::baseline:
+                {
+                    pm.addPass(hl::createHLLowerTypesPass());
```
xlauko, line 42: group shared passes between pipelines so we don't repeat ourself:
   - simplify-hl: `HLLowerTypes`, `DCE`, `LowerTypeDefs`
   - abi
   - to-ll
   - to-llvm
 and let `baseline` and `with_abi` invoke these pipelines?

lib/vast/Conversion/ABI/EmitABI.cpp
lkorenc: Will add comment, but this probably will not be classic `zip` in the end once we start supporting other `abi` classifications (`sret` for example).

lib/vast/Conversion/Common/IRsToLLVM.cpp
lkorenc: `Adaptor` does not have that method.

xlauko (approved):

PR #383, opened 2023-08-23
Add check for the decl in lookup table

kumarak:
Changelog:

- Emit an error if the Var decl is not present in the lookup table. During code generation,  while lowering `DeclRefExpr`, it searches for the declaration in the lookup table that may not be present. This situation could arise if the Decl node is not visited before. A potential solution could be visiting the Decl node if it is not present in the lookup table.

- Check for the null result type before setting the type in Unsupported Stmt Dialect

### DIFF
--- a/include/vast/CodeGen/CodeGenStmtVisitor.hpp
+++ b/include/vast/CodeGen/CodeGenStmtVisitor.hpp
@@ -551,9 +551,15 @@ namespace vast::cg {
 
         Operation* VisitEnumDeclRefExpr(const clang::DeclRefExpr *expr) {
             auto decl = clang::cast< clang::EnumConstantDecl >(expr->getDecl()->getUnderlyingDecl());
-            auto val = context().enumconsts.lookup(decl);
-            auto rty = visit(expr->getType());
-            return make< hl::EnumRefOp >(meta_location(expr), rty, val.getName());
+            if (auto val = context().enumconsts.lookup(decl)) {
+                auto rty = visit(expr->getType());
+                return make< hl::EnumRefOp >(meta_location(expr), rty, val.getName());
+            }
+
+            // Ref: https://github.com/trailofbits/vast/issues/384
+            // github issue to avoid emitting error if declaration is missing
+            context().error("error: missing enum constant declaration " + decl->getName());
+            return nullptr;
         }
 
         Operation* VisitVarDeclRefExprImpl(const clang::DeclRefExpr *expr, Value var) {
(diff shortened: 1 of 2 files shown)


include/vast/CodeGen/CodeGenStmtVisitor.hpp
```diff
@@ -563,11 +569,24 @@ namespace vast::cg {
...
+
+            // Ref: https://github.com/trailofbits/vast/issues/384
+            // github issue to avoid emitting error if declaration is missing
+            context().error("error: missing variable declaration " + decl->getName());
+            return nullptr;
         }
 
         Operation* VisitFileVarDeclRefExpr(const clang::DeclRefExpr *expr) {
             auto decl = getDeclForVarRef(expr);
+            if (auto file_var = context().vars.lookup(decl); !file_var) {
```
lkorenc, line 584:
```suggestion
            if (!context().vars.lookup(decl)) {
```

lkorenc (commented):
Fix commit messages please.

lib/vast/Dialect/Unsupported/UnsupportedOps.cpp
```diff
@@ -25,15 +25,18 @@ namespace vast::unsup {
...
 
     void UnsupportedStmt::build(
-        Builder &bld, State &st, llvm::StringRef name, Type rty, const std::vector< BuilderCallBackFn > &builders
+        Builder &bld, State &st, llvm::StringRef name, Type rty,
+        const std::vector< BuilderCallBackFn > &builders
     ) {
         InsertionGuard guard(bld);
-        st.addTypes(rty);
+        // Optional, add a check if rty exist.
+        if (rty) {
```
xlauko, line 33: Does it break anything when we store empty type? It should mean nullopt as intended in such case.

xlauko (approved):

lib/vast/Dialect/Unsupported/UnsupportedOps.cpp
kumarak: @xlauko, Yes if the type is null, it asserts inside the llvm mlir code that checks if the type is valid.

Jezurko:
Do you have some examples of when this happens?
I have been doing some work on fixing unvisited `VarDecl`s in #368

kumarak:
> Do you have some examples of when this happens? I have been doing some work on fixing unvisited `VarDecl`s in #368

@Jezurko, The example you have mentioned in the PR will be one such case where incase of non-odr initialization the VarDecl will not be there in the lookup table. 
There are other cases as well where the file scoped variable was missing from the lookup table. I tried reproducing the issue with the small example but most of them worked and the issue only appeared when I was lowering AST's from the lua and curl.

PR #393, opened 2023-09-13
Add void constant for return

Jezurko:


lib/vast/Conversion/Common/IRsToLLVM.cpp
```diff
@@ -426,10 +426,33 @@ namespace vast::conv::irstollvm
 
         }
 
+        logical_result handle_void_const(
+                hl::ConstantOp op, conversion_rewriter &rewriter) const
+        {
+            auto result = op.getResult();
+            if (result.hasOneUse()) {
+                auto user = result.getUses().begin()->getOwner();
+                if (is_one_of_mlir<hl::ReturnOp, ll::ReturnOp, core::ImplicitReturnOp >(
```
xlauko, line 435: we should have ReturnTrait for this ... or at least a typelist at a single location

lib/vast/Dialect/HighLevel/Transforms/HLCanonize.cpp
```diff
@@ -0,0 +1,90 @@
...
+            rewriter->create< core::ImplicitReturnOp >(op.getLoc(), void_const.getResult());
+        }
+
+        void run(Operation *op, rewriter_t &rewriter) {
+            if (mlir::isa< hl::SkipStmt >(op)) {
+                to_remove.emplace_back(op);
+                return;
+            }
+
+            if (auto fn = mlir::dyn_cast< hl::FuncOp >(op)) {
```
xlauko, line 45: too complex, extract to separate function

xlauko (approved):

PR #411, opened 2023-10-11
Attribute visitor improvements

Jezurko:
- support for additional attributes (I covered all attributes that are used for `malloc` in `stdlib`)
- deduplication of attributes - I believe this shouldn't be necessary, but it might be a bug in the AST…

### DIFF
--- a/include/vast/CodeGen/CodeGenAttrVisitor.hpp
+++ b/include/vast/CodeGen/CodeGenAttrVisitor.hpp
@@ -81,12 +81,8 @@ namespace vast::cg {
         }
 
         mlir_attr VisitAllocSizeAttr(const clang::AllocSizeAttr *attr) {
-            auto size = attr->getElemSizeParam();
-            auto num = attr->getNumElemsParam();
-            if (num.isValid()) {
-                return make< hl::AllocSizeAttr >(size.getSourceIndex(), num.getSourceIndex());
-            }
-            return make< hl::AllocSizeAttr >(size.getSourceIndex(), int());
+            auto num = attr->getNumElemsParam().isValid() ? attr->getNumElemsParam().getSourceIndex() : int();
+            return make< hl::AllocSizeAttr >(attr->getElemSizeParam().getSourceIndex(), num);
         }
     };
 } // namespace vast::cg
--- a/include/vast/CodeGen/CodeGenAttrVisitor.hpp
+++ b/include/vast/CodeGen/CodeGenAttrVisitor.hpp
@@ -13,6 +13,8 @@ VAST_UNRELAX_WARNINGS
 
 #include <mlir/IR/Attributes.h>
 #include <vast/CodeGen/Types.hpp>
+#include <vast/CodeGen/CodeGenVisitorLens.hpp>
+#include <vast/CodeGen/CodeGenBuilder.hpp>
 
 namespace vast::cg {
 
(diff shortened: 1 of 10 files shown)


include/vast/CodeGen/CodeGenAttrVisitor.hpp
```diff
@@ -65,5 +65,26 @@ namespace vast::cg {
         mlir_attr VisitWarnUnusedResultAttr(const clang::WarnUnusedResultAttr *attr) {
             return make< hl::WarnUnusedResultAttr >();
         }
+
+        mlir_attr VisitRestrictAttr(const clang::RestrictAttr *attr) {
+            return make< hl::MallocAttr >();
```
xlauko, line 70: is this typo? why not `RestrictAttr`?

include/vast/CodeGen/CodeGenAttrVisitor.hpp
```diff
@@ -65,5 +65,26 @@ namespace vast::cg {
...
+            return make< hl::NoThrowAttr >();
+        }
+
+        mlir_attr VisitBuiltinAttr(const clang::BuiltinAttr *attr) {
+            return make< hl::BuiltinAttr >(attr->getID());
+        }
+
+        mlir_attr VisitAllocSizeAttr(const clang::AllocSizeAttr *attr) {
+            auto size = attr->getElemSizeParam();
+            auto num = attr->getNumElemsParam();
```
xlauko, line 83: Use `num = attr->getNumElemsParam().isValid() ? ...`
and only a single `hl::AllocSizeAttr` constructor.

Maybe, i deally monadic builder with binds.

include/vast/CodeGen/Util.hpp
```diff
@@ -17,7 +17,7 @@ namespace vast::cg
     template< typename T >
     gap::generator< T * > filter(auto from) {
         for (auto x : from) {
-            if (auto s = clang::dyn_cast< T >(x))
+            if (auto s = mlir::dyn_cast< T >(x))
```
xlauko, line 20: what if we remove namespace completely? ADL should be more general here?

include/vast/CodeGen/CodeGenAttrVisitor.hpp
Jezurko: Nope, `RestrictAttr` is the name Clang uses for the `malloc` attribute… And it's even listed as such in their [docs](https://clang.llvm.org/docs/AttributeReference.html#malloc).
I named it malloc since that's how I discovered it in the source code I was using and the spelling of the attribute there is `malloc`. I can change it if you think it will be better.

xlauko: let's keep clangs naming for consistency

include/vast/CodeGen/Util.hpp
```diff
@@ -8,17 +8,29 @@ VAST_RELAX_WARNINGS
...
     gap::generator< T * > filter(auto from) {
         for (auto x : from) {
-            if (auto s = clang::dyn_cast< T >(x))
+            if (auto s = dyn_cast< T >(x))
                 co_yield s;
         }
     }
+
+    template< typename list >
+    gap::generator< clang::Attr * >exclude_attrs(auto from) {
```
xlauko, line 29: space

include/vast/CodeGen/Util.hpp
```diff
@@ -8,17 +8,29 @@ VAST_RELAX_WARNINGS
...
+        for (auto &attr : from) {
```
xlauko, line 30: why ref when its pointer?

xlauko (approved):

PR #420, opened 2023-10-19
Fix conversion of chained logical operators

Jezurko:
`a && (!b || !c)`

### DIFF
--- a/lib/vast/Conversion/Core/ToLLVM.cpp
+++ b/lib/vast/Conversion/Core/ToLLVM.cpp
@@ -10,6 +10,8 @@ VAST_RELAX_WARNINGS
 #include <mlir/IR/IRMapping.h>
 VAST_UNRELAX_WARNINGS
 
+#include <iterator>
+
 #include "../PassesDetails.hpp"
 #include "vast/Conversion/Common/Passes.hpp"
 #include "vast/Conversion/Common/Patterns.hpp"
@@ -45,8 +47,10 @@ namespace vast
 
                 auto &first_block = lazy_region.front();
                 // The rewriter API doesn't provide a call to insert into a selected block
+                auto target_it = target->getIterator();
+                std::advance(target_it, 1);
                 rewriter.inlineRegionBefore(
-                    lazy_region, *target->getParent(), ++(target->getIterator())
+                    lazy_region, *target->getParent(), target_it
                 );
                 rewriter.mergeBlocks(&first_block, target, std::nullopt);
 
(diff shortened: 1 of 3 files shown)


lib/vast/Conversion/Core/ToLLVM.cpp
```diff
@@ -45,8 +47,10 @@ namespace vast
 
                 auto &first_block = lazy_region.front();
                 // The rewriter API doesn't provide a call to insert into a selected block
+                auto target_it = target->getIterator();
+                std::advance(target_it, 1);
```
xlauko, line 51: use directly 

```
                auto target_it = std::next(target->getIterator());
```

lib/vast/Conversion/Core/ToLLVM.cpp
```diff
@@ -104,7 +108,12 @@ namespace vast
 
                 cond_br_lhs(rewriter, op.getLoc(), cmp_lhs, rhs_block, end_block);
 
-                rewriter.setInsertionPointToEnd(rhs_block);
+                // In the case that `rhs` consists of multiple blocks we need to
+                // insert to the last block…
```
xlauko, line 112: ...?

lib/vast/Conversion/Core/ToLLVM.cpp
```diff
@@ -104,7 +108,12 @@ namespace vast
...
+                // e.g.: in a && (b || c) mlir first matches the `||` which creates
+                // it's own cf blocks and then matches the `&&`
+                auto rhs_end_it = std::prev(end_block->getIterator(), 1);
```
xlauko, line 115: no need of `1` its implicit argument of `std::prev`

lib/vast/Conversion/Core/ToLLVM.cpp
```diff
@@ -104,7 +108,12 @@ namespace vast
...
+                rewriter.setInsertionPointToEnd(&*rhs_end_it);
```
xlauko, line 116: does this match [getSinglePredecessor](https://mlir.llvm.org/doxygen/classmlir_1_1Block.html#a72289303038da4c089aeb47ca97c29ac)?

lib/vast/Conversion/Core/ToLLVM.cpp
Jezurko: I tried to update the comment. Is it now explained better or should I try more?

xlauko (approved):

PR #449, opened 2023-11-16
Handle abi alignment when lowering structures through ABI passes.

lkorenc:


include/vast/Conversion/ABI/AggregateTypes.hpp
```diff
@@ -44,6 +44,15 @@ namespace vast::conv::abi
...
                 offset = 0;
             }
+
+
+            std::optional< std::size_t > needs_align_paddding(mlir_type type)
+            {
+                auto align = parent.dl.getTypeABIAlignment(type) * 8;
+                if (offset % align != 0)
+                    return { align - (offset % align) };
+                return {};
```
xlauko, line 54: why not just return zero padding instead of unnecessary allocation of optional?

include/vast/Conversion/ABI/AggregateTypes.hpp
```diff
@@ -99,6 +108,12 @@ namespace vast::conv::abi
                         this->abi_op.getLoc(), type, arg(),
                         start, this->offset);
             };
+
+            void adjust_by_align(mlir_type type)
+            {
+                if (auto pad_by = this->needs_align_paddding(type))
```
xlauko, line 114: if it returns zero, no need to condition here, just add padding to offset

include/vast/Dialect/HighLevel/HighLevelUtils.hpp
```diff
@@ -102,19 +110,25 @@ namespace vast::hl
...
+        }
+    }
 
+    // Given record `root` emit `hl::RecordMemberOp` casted as rvalue for each
+    // its member.
+    auto generate_values_of_record_members(operation root, auto &bld)
+        -> gap::generator< hl::ImplicitCastOp >
+    {
+        for (auto member_ptr : generate_ptrs_to_members(root, bld)) {
+            co_yield implicit_cast_lvalue_to_rvalue(bld, root->getLoc(), member_ptr);
```
xlauko, line 126: you want to take member_ptr location instead of root here?

lib/vast/Conversion/Common/Common.hpp
```diff
@@ -53,6 +53,14 @@ namespace vast::conv::irstollvm {
                 loc, index_type, rewriter.getIntegerAttr(index_type, idx)
             );
         }
+
+        // ZExt when source and target type are same is a hard error.
+        static mlir::Value maybe_zext(auto &rewriter, auto loc, mlir_type res_type, auto val) {
```
xlauko, line 58: is this ever used? there is no use in this PR

lib/vast/Conversion/Common/IRsToLLVM.cpp
```diff
@@ -104,37 +104,14 @@ namespace vast::conv::irstollvm
                 return arg;
```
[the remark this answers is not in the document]
lkorenc: Well they are direct result of the extract being lowered, so imho they should have its location?

xlauko (approved):

PR #458, opened 2023-12-07
Implement configurable codegen pipelines

xlauko:


include/vast/Util/Pipeline.hpp
```diff
@@ -0,0 +1,187 @@
...
+
+
+    using pipeline_step_builder = std::function< pipeline_step_ptr(void) >;
+
+    //
+    // initilizer wrapper to setup dependencies after make is called
+    //
+    template< typename step_t >
+    struct pipeline_step_init : pipeline_step_ptr {
+        using pipeline_step_ptr::pipeline_step_ptr;
```
lkorenc, line 69: Is this even doing anything when we have a `args_t && ...` ctor later? (I think it will catch all invocations).

lkorenc (approved):

PR #466, opened 2023-12-12
cg: Deduplicate variable initialization content and typeof defining ops

Jezurko:


include/vast/CodeGen/CodeGenDeclVisitor.hpp
```diff
@@ -533,12 +533,15 @@ namespace vast::cg {
             }).getDefiningOp();
 
             if (decl->hasInit()) {
-                auto guard = insertion_guard();
                 auto declared = mlir::dyn_cast< hl::VarDeclOp >(var_decl);
-                set_insertion_point_to_start(&declared.getInitializer());
+                auto &initializer = declared.getInitializer();
+                if (initializer.op_begin() == initializer.op_end()) {
```
xlauko, line 538: why is this check here?

Jezurko: To protect the initialization region from begin filled up multiple times (when declare() returns a variable found in the hash maps instead of creating a new one). I have made it more reasonable and added a short explanation.

xlauko (approved):

PR #467, opened 2023-12-13
Remove core strlit

Jezurko:
Custom string literal is no longer required and is an unnecessary duplication of logic already provided by MLIR.


include/vast/Conversion/TypeConverters/TypeConverter.hpp
```diff
@@ -167,4 +167,13 @@ namespace vast::conv::tc {
                 .template take_wrapped< maybe_attr_t >();
         };
     }
+
+    auto convert_string_attr(auto &type_converter) {
+        return [&type_converter](mlir::StringAttr attr) -> maybe_attr_t {
+            return Maybe(attr.getType())
+                .and_then([&](auto type) { return type_converter.convertType(type); })
```
lkorenc, line 174: Does this not have some getter? If not maybe it should.

lkorenc (approved):

PR #490, opened 2024-01-29
Fix lowering of global variables to llvm

Jezurko:
- [ ] Add conversion for global variables to llvm
- [ ] Check that the example program compiles
```
#include <stdint.h>
#undef KEY
#if defined(__i386)
# define KEY '_','_','i','3','8','6'
#elif defined(__x86_64)
# define KEY '_','_','x','8','6','_','6','4'
#elif defined(__PPC64__)
# define KEY '_','_','P','P','C','6','4','_','_'
#elif defined(__ppc64__)
# define KEY '_','_','p','p','c','6','4','_','_'
#elif defined(__PPC__)
# define KEY '_','_','P','P','C','_','_'
#elif defined(__ppc__)
# define KEY '_','_','p','p','c','_','_'
#elif defined(__aarch64__)
# define KEY '_','_','a','a','r','c','h','6','4','_','_'
#elif defined(__ARM_ARCH_7A__)
# define KEY '_','_','A','R','M','_','A','R','C','H','_','7','A','_','_'
#elif defined(__ARM_ARCH_7S__)
# define KEY '_','_','A','R','M','_','A','R','C','H','_','7','S','_','_'
#endif

#define SIZE (sizeof(unsigned short))
static char info_size[] =  {'I', 'N', 'F', 'O', ':', 's','i','z','e','[',
  ('0' + ((SIZE / 10000)%10)),
  ('0' + ((SIZE / 1000)%10)),
  ('0' + ((SIZE / 100)%10)),
  ('0' + ((SIZE / 10)%10)),
  ('0' +  (SIZE    % 10)),
  ']',
#ifdef KEY
  ' ','k','e','y','[', KEY, ']',
#endif
  '\0'};

#ifdef __CLASSIC_C__
int main(argc, argv) int argc; char *argv[];
#else
int main(int argc, char *argv[])
#endif
{
  int require = 0;
  require += info_size[argc];
  (void)argv;
  return require;
}
```

### DIFF
--- a/lib/vast/Conversion/Common/IRsToLLVM.cpp
+++ b/lib/vast/Conversion/Common/IRsToLLVM.cpp
@@ -12,6 +12,7 @@ VAST_RELAX_WARNINGS
 #include <mlir/Transforms/GreedyPatternRewriteDriver.h>
 
 #include <llvm/ADT/APFloat.h>
+#include <llvm/ADT/SmallVector.h>
 VAST_UNRELAX_WARNINGS
 
 #include "../PassesDetails.hpp"
@@ -373,7 +374,14 @@ namespace vast::conv::irstollvm
 
             // Sadly, we cannot build `mlir::LLVM::GlobalOp` without
             // providing a value attribute.
-            auto dummy_value = rewriter.getIntegerAttr(target_type, 0);
+            auto create_dummy_value = [&] () -> mlir::Attribute {
+                if (auto trg_arr = mlir::dyn_cast< mlir::LLVM::LLVMArrayType >(target_type)) {
+                    attrs_t arr(trg_arr.getNumElements(),
+                                rewriter.getIntegerAttr(trg_arr.getElementType(), 0));
+                    return rewriter.getArrayAttr(arr);
+                }
+                return rewriter.getIntegerAttr(target_type, 0);
+            };
 
             // So we know this is a global, otherwise it would be in `ll:`.
             auto gop = rewriter.create< mlir::LLVM::GlobalOp >(
@@ -382,7 +390,7 @@ namespace vast::conv::irstollvm
                     // TODO(conv:irstollvm): Constant.
                     true,
                     LLVM::Linkage::Internal,
-                    op.getName(), dummy_value);
+                    op.getName(), create_dummy_value());
 
             // If we want the global to have a body it cannot have value attribute.
             gop.removeValueAttr();
--- a/lib/vast/Conversion/Common/IRsToLLVM.cpp
+++ b/lib/vast/Conversion/Common/IRsToLLVM.cpp
@@ -426,10 +426,10 @@ namespace vast::conv::irstollvm
             auto create_dummy_value = [&] () -> mlir::Attribute {
                 if (auto trg_arr = mlir::dyn_cast< mlir::LLVM::LLVMArrayType >(target_type)) {
                     attrs_t arr(trg_arr.getNumElements(),
-                                rewriter.getIntegerAttr(trg_arr.getElementType(), 0));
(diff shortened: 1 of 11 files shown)


include/vast/Dialect/HighLevel/HighLevelOps.td
```diff
@@ -1035,7 +1035,7 @@ def AddrLabelExpr
 }
 
 def InitListExpr
-  : HighLevel_Op< "initlist" >
+  : HighLevel_Op< "initlist", [NoMemoryEffect] >
```
xlauko, line 1038: is that true? what if I have `{ f(), g() }`?

lib/vast/Conversion/Common/IRsToLLVM.cpp
```diff
@@ -48,6 +49,54 @@ namespace vast::conv::irstollvm
...
 
+    template< typename self_t >
+    struct value_builder
+    {
+      private:
+        const self_t &self() const { return static_cast< const self_t & >(*this); }
+
+      protected:
+
+        mlir::Value construct_value(auto &rewriter, mlir_value val) const {
```
xlauko, line 60: `mlir::Value` vs `mlir_value`, use one consistently at least localy

lib/vast/Conversion/Common/IRsToLLVM.cpp
```diff
@@ -48,6 +49,54 @@ namespace vast::conv::irstollvm
...
+            auto op = val.getDefiningOp();
+            VAST_ASSERT(op && op->getResults().size() == 1);
+            auto trg_type = self().convert(op->getResults()[0].getType());
+            return construct_value(rewriter, op, trg_type);
+        }
+
+        mlir::Value construct_value(auto &rewriter, operation op, mlir_type type) const {
+            auto init_list = mlir::dyn_cast< hl::InitListExpr >(op);
+            if (!init_list)
```
xlauko, line 69: is only init list multi-value constructor?

lib/vast/Conversion/Common/IRsToLLVM.cpp
```diff
@@ -48,6 +49,54 @@ namespace vast::conv::irstollvm
...
+        }
+
+        mlir::Value construct_aggregate_value(
+            conversion_rewriter &rewriter, hl::InitListExpr init_list,
+            mlir_type aggragate_type
+        ) const {
+            // Currently we are only supporting this type of initialization so
+            // better be defensive about it.
+            mlir_value init = self().undef(rewriter, init_list.getLoc(), aggragate_type);
+            for (auto [idx, o] : llvm::enumerate(init_list.getElements()))
```
xlauko, line 88: what is `o` give it more meaningful abbreviation?

xlauko (approved):

lib/vast/Conversion/Common/IRsToLLVM.cpp
lkorenc: `operand`, I think. Will do.

include/vast/Dialect/HighLevel/HighLevelOps.td
lkorenc: @Jezurko ^

Jezurko: Yeah, this should be removed. I added it at first, because without it was triggering some asserts in some later passes (but it shouldn't get there anyway).

lib/vast/Conversion/Common/IRsToLLVM.cpp
lkorenc: For now this is the only one we expect. If some new shows we will adjust, I am just being defensive.

xlauko:
@Jezurko @lkorenc can you add example program from this issue to tests too.

PR #491, opened 2024-01-30
Allow to disable pipelines by name.

xlauko:
Introduces set of options  `-vast-disable-<pipeline-name>` which allow the user to turn off specific steps of pipeline.

Adds capability to name pipelines and rename passes.

### DIFF
--- a/cmake/AddVAST.cmake
+++ b/cmake/AddVAST.cmake
@@ -59,6 +59,15 @@ function(vast_tablegen ofn)
         )
     endif()
 
+    if (CMAKE_GENERATOR MATCHES "Visual Studio")
+        # Visual Studio has problems with llvm-tblgen's native --write-if-changed
+        # behavior. Since it doesn't do restart optimizations anyway, just don't
+        # pass --write-if-changed there.
+        set(vast_tblgen_change_flag)
+    else()
+        set(vast_tblgen_change_flag "--write-if-changed")
+    endif()
+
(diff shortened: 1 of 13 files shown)


cmake/AddVAST.cmake
```diff
@@ -75,19 +84,20 @@ function(vast_tablegen ofn)
     list(REMOVE_ITEM vast_tablegen_includes "")
     list(TRANSFORM vast_tablegen_includes PREPEND -I)
 
-    set(tablegen_exe ${VAST_TABLEGEN_EXE})
-    set(tablegen_depends ${VAST_TABLEGEN_TARGET} ${tablegen_exe})
-
+    set(vast_tablegen_exe ${VAST_TABLEGEN_EXE})
+    set(vast_tablegen_depends ${VAST_TABLEGEN_TARGET} ${vast_tablegen_exe})
+
```
lkorenc, line 89: Looks like this introduces trailing whitespace?

lib/vast/Frontend/Pipelines.cpp
```diff
@@ -25,7 +16,7 @@ namespace vast::cc {
 
         // Simplifies high level MLIR
         pipeline_step_ptr reduce_high_level() {
-            return compose(
+            return compose( "reduce-hl",
```
lkorenc, line 19:
```suggestion
            return compose("reduce-hl",
```

lkorenc (approved):

Issue #388, opened 2023-09-05
Missing terminator in empty `llvm` functions.


xlauko:
On the following program:

```
void foo() {}
```

`vast-front -vast-emit-mlir=llvm` dies with:

```
empty block: expect at least a terminator
```

xlauko:
Related to #315

lkorenc:
Resolved.

PR #492, opened 2024-02-01
Fix minor issues in ABI conversions.

### LINKED ISSUE #376: Support abi conversion of declarations.
Declarations are currently skipped, since I was running into errors when trying to figure out location information for arguments. This should not be a hard thing to finish implementing - basically only the type needs to be changes, there is no need to emit prologues or epilogues as the function is without body.

lkorenc:
closes #376 and fixes incorrect name prefix during abi conversions.

### DIFF
--- a/lib/vast/Conversion/ABI/EmitABI.cpp
+++ b/lib/vast/Conversion/ABI/EmitABI.cpp
@@ -830,7 +830,7 @@ namespace vast
             auto is_return_legal = [&](hl::ReturnOp op)
             {
                 auto func = op->getParentOfType< abi::FuncOp >();
-                if (!func || func.getName() == "main")
+                if (!func)
                     return true;
 
                 for (auto val : op.getResult())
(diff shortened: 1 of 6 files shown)


test/vast/Compile/SingleSource/puts-a.c
```diff
@@ -1,5 +1,4 @@
-// RUN: %vast-front -o %t %s && %t hello | %file-check %s
```
xlauko, line 1: `-vast-pipeline=with-abi` should not exists anymore?

lkorenc:
close #388

lib/vast/Conversion/ABI/EmitABI.cpp
```diff
@@ -227,7 +227,7 @@ namespace vast
                         op.getLoc(),
                         // Temporal, to avoid verification issues, will be changed once
                         // original func is removed.
-                        "vast.abi" + op.getName().str(),
+                        "vast.abi." + op.getName().str(),
```
xlauko, line 230: make this some global named constant `vast::abi_prefix`, it is already on two places

xlauko (approved):

PR #494, opened 2024-02-02
Fix memory corruption related to unregistered linkage attribute.

xlauko:


include/vast/Dialect/Core/CoreAttributes.td
```diff
@@ -120,4 +132,6 @@ def SourceLanguageAttr : EnumAttr< Core_Dialect, SourceLanguage, "lang" > {
   }];
 }
 
+include "vast/Dialect/Core/Linkage.td"
```
Jezurko, line 135: We already have this include in `vast/Dialect/Core/Core.td`. Maybe we could have both includes at places where both attributes and linkage are required? Or if we consider these a single package, then we should just remove it from the "main" dialect file, just for clarity.

Jezurko (approved):

PR #498, opened 2024-02-05
Change getFunctionType() to return an optional

PappasBrent:
- Change each `getFunctionType()` overload to return an optional `FunctionType` so that they can return an empty option if they fail instead of crashing VAST.
- Check the return value of `getFunctionType()` in `VisitIndirectCall()` and return an `IndirectCallOp` only if the return value is not an empty option. Otherwise, return an empty op for a fallback visitor (e.g., the unsupported visitor) to handle.

### DIFF
--- a/include/vast/CodeGen/CodeGenStmtVisitor.hpp
+++ b/include/vast/CodeGen/CodeGenStmtVisitor.hpp
@@ -978,8 +978,14 @@ namespace vast::cg {
         operation VisitIndirectCall(const clang::CallExpr *expr) {
             auto callee = VisitIndirectCallee(expr->getCallee())->getResult(0);
             auto args   = VisitArguments(expr);
-            auto type   = hl::getFunctionType(callee.getType(), context().mod.get()).getResults();
-            return make< hl::IndirectCallOp >(meta_location(expr), type, callee, args);
+            auto type   = hl::getFunctionType(callee.getType(), context().mod.get());
+            if (type) {
+                return make< hl::IndirectCallOp >(
+                    meta_location(expr), type.getResults(), callee, args
+                );
+            }
+
+            return {};
         }
 
         operation VisitCallExpr(const clang::CallExpr *expr) {
--- a/lib/vast/Dialect/HighLevel/HighLevelTypes.cpp
+++ b/lib/vast/Dialect/HighLevel/HighLevelTypes.cpp
@@ -92,6 +92,10 @@ namespace vast::hl
     }
 
     core::FunctionType getFunctionType(mlir::CallInterfaceCallable callee, vast_module mod) {
+        if (!callee) {
+            return {};
+        }
+
         if (auto sym = callee.dyn_cast< mlir::SymbolRefAttr >()) {
             return mlir::dyn_cast_or_null< FuncOp >(
                 mlir::SymbolTable::lookupSymbolIn(mod, sym)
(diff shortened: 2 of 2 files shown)


lkorenc:
Clear up commits please, there are changes to both `hl` and `cg`. The message also does not follow the guidelines we use.

xlauko:
There is no need to return optional, you can just return empty type `{}`, it behaves nicely in conditionals and does not require additional indirection and allocation.

xlauko:
Also this is not solving the root cause why did we not manage to create FunctionType? Do you have source? There will be just missing case in the type resolver. Unsupported representation is kinda useless still.

PappasBrent:
> Do you have source?

Yes, vast crashes when I use the default_stmt_visitor in Macroni to lower [net/mac80211/tx.c](https://github.com/torvalds/linux/blob/54be6c6c5ae8e0d93a6c4641cb7528eb0b6ba478/net/mac80211/tx.c) in the Linux kernel.
I will create an issue for this error once I can create a minimal reproducible example using `vast-front`.

> Unsupported representation is kinda useless still.

It is useful for Macroni since it enables Macroni to perform macro analyses on at least part of the AST, though I agree it does not solve the problem as to why vast could not create a FunctionType.

PappasBrent:
I'm closing this PR until I can create a minimal working example of the issue it addresses.

xlauko:
You can definitely keep this open, and ideally replace optional with `{}` we can merge it then. :) 
UNREACHABLES are there from times we did not have unsupported visitor. 

Also I plan to add option to turn off unsupported visitor, so we will catch these kind of errors in the visitor.

When you have a minimal example just create an issue.

pgoodman:
@xlauko Perhaps you could make an `UnreachableVisitor`.

PappasBrent:
Got it, thank you for your patience

lkorenc:
Changes look good, but please clean up the commits (there is no need to have those original attempts with `optional` in the tree).

include/vast/CodeGen/CodeGenStmtVisitor.hpp
```diff
@@ -978,8 +978,14 @@ namespace vast::cg {
...
-            auto type   = hl::getFunctionType(callee.getType(), context().mod.get()).getResults();
-            return make< hl::IndirectCallOp >(meta_location(expr), type, callee, args);
+            auto type   = hl::getFunctionType(callee.getType(), context().mod.get());
+            if (type) {
+                return make< hl::IndirectCallOp >(
+                    meta_location(expr), type.getResults(), callee, args
+                );
+            }
+
+            return {};
```
pgoodman, line 988: @PappasBrent @xlauko Under what situations is getting a function type here not possible? Can it be documented here in the code with a comment? What about getting the type from `expr->getCallee()`?

Should there be a TODO/FIXME comment?

pgoodman: Is this a C++-ism, e.g. calling an `operator()`, or is it related to the issue of builtins? E.g. how Clang has a `BuiltinFn` type (which is super unhelpful and annoying).

xlauko: Thats literally what was there as `VAST_UNIMPEMENTED` and what @PappasBrent was changing in this PR. I don't know every case when this fails but now this is deferred to visitor to yield unsupported type or an error message.

xlauko (approved):

include/vast/CodeGen/CodeGenStmtVisitor.hpp
pgoodman: I mean a comment describing the case that was hit, so a reader of the code isn't surprised by the fact that one can't always get a function type. I think the code should document the "surprises" of the unhappy path when possible.

PR #522, opened 2024-03-06
Unused Definition Elimination

xlauko:


lib/vast/Dialect/HighLevel/Transforms/UDE.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+#if !defined(NDEBUG)
+    constexpr bool debug_ude_pass = false;
+    #define VAST_UDE_DEBUG(...) VAST_REPORT_WITH_PREFIX_IF(debug_ude_pass, "[UDE] ", __VA_ARGS__)
+#else
+    #define VAST_UDE_DEBUG(...)
+#endif
+
+    template< typename ... args_t >
+    bool is_one_of( operation op ) { return ( mlir::isa< args_t >( op ) || ... ); }
```
lkorenc, line 37: `mlir::isa<>` already accepts more types.

lib/vast/Dialect/HighLevel/Transforms/UDE.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+#if !defined(NDEBUG)
+    constexpr bool debug_ude_pass = false;
+    #define VAST_UDE_DEBUG(...) VAST_REPORT_WITH_PREFIX_IF(debug_ude_pass, "[UDE] ", __VA_ARGS__)
+#else
+    #define VAST_UDE_DEBUG(...)
+#endif
+
+    constexpr auto keep_only_if_used = [] (auto, auto) { /* do not keep by default */ return false; };
+
```
Jezurko, line 37: I don't understand the purpose of this lambda. Why is it preferred over simply returning `false` from the caller?

PR #541, opened 2024-04-08
Add sv-comp CI run

Jezurko:


.github/workflows/prerelease.yml
```diff
@@ -63,7 +70,239 @@ jobs:
...
+          -C=../cmake/caches/O3.cmake \
+          -DTEST_SUITE_COLLECT_CODE_SIZE=OFF \
+          -DTEST_SUITE_SUBDIRS=SingleSource \
+          -DVAST_TEST_SUITE_ENABLE=ON \
+          -DVAST_TEST_SUITE_TARGET=${{ matrix.vast-target }}
+        make -j $(nproc) -k
+
+     - name: Collect run data (mlir)
+       if: matrix.vast-target != 'bin'
+       working-directory: ./llvm-test-suite/build
```
xlauko, line 133: why in `build` dir when everything you access is in `../`?

.github/workflows/prerelease.yml
```diff
@@ -63,7 +70,239 @@ jobs:
...
+       with:
+         name: VAST
+
+     - name: Unpack VAST
+       run: mkdir vast && tar -xf VAST-* -C vast --strip-components=1
+
+     - name: Export vast binaries
+       run: echo "${PWD}/vast/bin/" >> $GITHUB_PATH
+
+     - name: Install test suite dependencies
```
xlauko, line 218: I think `git` must be in the container already

xlauko (approved):

.github/workflows/prerelease.yml
Jezurko: The script performs globbing on the results and generates the benchmark names on that. So it would generate incorrect `.json` with different names than expected.

PR #542, opened 2024-04-12
Constant folding facade

xlauko:
Make vast-front and vast-opt be able to constant fold high-level dialect.

- adds `-vast-canonicalize` option.
- adds facade for binary operations folding and example of `add` folder
- implements high-level constant materializer
- introduces `ConstantLikeAttrTrait` to easily gather materializable constants


lib/vast/Dialect/HighLevel/HighLevelOps.cpp
```diff
@@ -41,8 +43,146 @@ namespace vast::hl
...
 
+
+    //===----------------------------------------------------------------------===//
+    // ArithBinOps
+    //===----------------------------------------------------------------------===//
+
+    FoldResult checked_int_arithmetic(
+        mlir_type type, auto adaptor, auto &&op
+    ) {
+        if (auto lhs = mlir::dyn_cast_or_null< core::IntegerAttr >(adaptor.getLhs())) {
```
lkorenc, line 54: I would do
```
auto rhs = ...
auto lhs = ...
if (!rhs || !lhs)
  return {};
return op(rhs, lhs);
```
I think it is in a sense cleaner as that `rhs/lhs` nesting is not really "dependent" on each step.

lkorenc (approved):

PR #543, opened 2024-04-16
LLVM 18 port

lkorenc:


lib/vast/Conversion/Core/Passes.cpp
```diff
@@ -18,7 +18,7 @@ namespace vast::conv::pipeline {
         // This is necessary to have line tables emitted and basic debugger
         // working. In the future we will add proper debug information emission
         // directly from our frontend.
-        return nested< mlir::LLVM::LLVMFuncOp >(mlir::LLVM::createDIScopeForLLVMFuncOpPass)
+        return nested< vast_module >(mlir::LLVM::createDIScopeForLLVMFuncOpPass)
```
xlauko, line 21: why this?

test/vast/Conversion/Common/CoreToLLVM/binland-a.c
```diff
@@ -1,4 +1,4 @@
-// RUN: %vast-cc1 -vast-emit-mlir=hl %s -o - | %vast-opt-core-to-llvm | %file-check %s
+// RUN: %vast-cc1 -vast-emit-mlir=hl %s -o - | %vast-opt-irs-to-llvm | %file-check %s
```
xlauko, line 1: why irs-to-llvm when it is in `CoreToLLVM` folder?

lkorenc: I will clean this when we update tests, really what that `%vast-opt-core-to-llvm` did was the same as the `irs-to-llvm` but just omitted some random passes in the middle.

lib/vast/Conversion/Core/Passes.cpp
lkorenc: Because upstream changed what the pass operates on (it is on module now) - the upstream commit message had some reasoning why this happened.

xlauko (approved):

PR #575, opened 2024-04-26
Enum lowering and option to take snapshots.

lkorenc:
Adds pass that lowers enum.

Adds `-vast-snapshot-at=pass1; ... ;passN` to `vast-front` that will store MLIR module after given pass. I plan to use this to streamline test definition and eliminate redundant pipeline redefinitions.


include/vast/Conversion/Passes.td
```diff
@@ -197,4 +197,16 @@ def HLToHLBI : Pass<"vast-hl-to-hl-builtin", "mlir::ModuleOp"> {
   ];
 }
 
+def LowerEnums : Pass<"vast-hl-lower-enums", "mlir::ModuleOp"> {
+  let summary = "Lower high level representation of enums.";
+  let description = [{
+    This pass is still a work in progress.
```
xlauko, line 203: Add description into what you lower enums

lib/vast/Conversion/FromHL/EmitLazyRegions.cpp
```diff
@@ -133,6 +133,7 @@ namespace vast
         static conversion_target create_conversion_target(mcontext_t &context) {
             conversion_target target(context);
             target.addLegalDialect< vast::core::CoreDialect >();
+            target.markUnknownOpDynamicallyLegal([](auto){ return true; });
```
xlauko, line 136: why is this needed now?

lib/vast/Conversion/FromHL/Passes.cpp
```diff
@@ -48,9 +48,14 @@ namespace vast::conv::pipeline {
...
     }
 
+
+    pipeline_step_ptr hl_lower_enums() {
+        return pass(createLowerEnumsPass);
+    }
     pipeline_step_ptr to_ll() {
         return compose( "to-ll",
             hl_to_ll_func,
+            hl_lower_enums,
```
xlauko, line 58: This should probably go to hl simplification pass, as it stays in `hl` no?

lib/vast/Frontend/Options.cpp
```diff
@@ -49,7 +49,6 @@ namespace vast::cc {
     std::optional< string_ref > vast_args::get_option(string_ref name) const {
         if (auto opt = detail::get_option_impl(args, name)) {
             if (auto [lhs, rhs] = opt->split('='); !rhs.empty()) {
-                VAST_ASSERT(!is_options_list(rhs));
```
xlauko, line 52: why did you removed this? if you want to use list option you should pick it using `get_options_list`

lib/vast/Conversion/FromHL/LowerEnums.cpp
```diff
@@ -0,0 +1,177 @@
...
+            // added (this is easier to debug),
+            // but eventually we probably should swap these to patterns.
+            replace_constants(enums);
+            replace_var_init(enums);
+            replace_types(enums);
+
+            erase_enums(enums);
+        }
+
+        void replace_constants(const enums_info &enums) {
```
xlauko, line 112: This should be cannonicalization pass (i.e. just definition how to emit constant from enum op) rather than pass.

lib/vast/Frontend/Options.cpp
lkorenc: Because for `option="pass1;pass2"` and call to `get_options_list` this will assert on `pass1;pass2` as this is used to parse `[lhs,rhs]`.

lib/vast/Conversion/FromHL/EmitLazyRegions.cpp
lkorenc: This should really always be present - otherwise if there is an op that is not explicitly specified by other callback (or is from some "external" dialect to the pass), it will be treated as illegal and conversion will fail even if no work was to be done for that op.

include/vast/Util/Pipeline.hpp
```diff
@@ -65,9 +66,77 @@ namespace vast {
...
+                false,                                // after change
+                true,                                 // after failure
+                os
+            );
+        }
+
         llvm::DenseSet< pass_id_t > seen;
     };
 
+    struct with_snapshots : mlir::PassInstrumentation
```
xlauko, line 83: Add documentation to `docs/Tools/vast-front.md`

xlauko: Also might be worth to mention in `docs/GettingStarted/debug.md` in `### Pipeline Information`.

test/vast/Conversion/enum-a.c
```diff
@@ -0,0 +1,37 @@
+// RUN: %vast-front -vast-snapshot-at="vast-hl-lower-enums" %s
+// RUN: echo "$(ls)"
```
xlauko, line 2: can be simplified as: 
```
// RUN: %file-check %s -input-file=$(basename %s .c).vast-hl-lower-enums -check-prefix=ENUM
```

xlauko (approved):

PR #584, opened 2024-05-13
`core::SelectOp` and some `void` related fixes.

lkorenc:
Adds lowering of `core::SelectOp`. Fixes some corner cases related to `void` "values" being returned.

### DIFF
--- a/lib/vast/Conversion/Core/ToLLVM.cpp
+++ b/lib/vast/Conversion/Core/ToLLVM.cpp
@@ -36,9 +36,19 @@ namespace vast
             using base::base;
 
             auto lazy_into_block(
-                Operation* lazy_op, Block* target, conversion_rewriter &rewriter) const
-            {
+                Operation* lazy_op, Block* target, conversion_rewriter &rewriter
+            ) const -> std::tuple< mlir_value, mlir::Block * > {
                 auto &lazy_region = mlir::dyn_cast< core::LazyOp >(*lazy_op).getLazy();
+
+                // In case `lazy_op` already has multiple blocks (nested lazy ops)
+                // we need to return the block that does not yet have the control flow.
+                // This will be the block with the yield, but if it is in the first block,
+                // it will get merged.
+                auto end_block = [&] {
+                    if (&lazy_region.front() == &lazy_region.back())
+                        return target;
+                    return &lazy_region.back();
+                }();
 
                 // Last block should have hl.value.yield with the final value
                 auto &yield = lazy_region.back().back();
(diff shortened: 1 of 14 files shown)


lib/vast/Conversion/ToLLVM/Common.hpp
```diff
@@ -77,6 +77,14 @@ namespace vast::conv::irstollvm {
...
             return convert(ptr.getElementType());
         }
+
+        std::vector< mlir::Value > strip_voids(const auto &values) const {
+            std::vector< mlir::Value > out;
+            for (auto v : values)
+                if (!mlir::isa< LLVM::LLVMVoidType >(v.getType()))
+                    out.push_back(v);
+            return out;
+        }
```
xlauko, line 87:
```
values_t filter_voids(const auto& values) const {
    values_t out;
    std::ranges::copy_if(values, std::back_inserter(out), [] (const mlir_value &v) {
        return !mlir::isa< LLVM::LLVMVoidType >(v.getType());
    });
    return out;
}
```

test/vast/Conversion/void-return-b.c
```diff
@@ -0,0 +1,8 @@
+// RUN: %vast-front -vast-emit-mlir=llvm -vast-snapshot-at="vast-irs-to-llvm" %s
+// RUN: %file-check %s -input-file=$(basename %s .c).vast-irs-to-llvm -check-prefix=LLVM
+
+// REQUIRES: hl.stmt.expr
```
xlauko, line 4: rename to better requires we support `hl.stmt.expr`

lib/vast/Conversion/Core/ToLLVM.cpp
```diff
@@ -127,9 +185,63 @@ namespace vast
...
+        struct select : lazy_base< core::SelectOp >
+        {
+            using op_t = core::SelectOp;
+            using base = lazy_base<  op_t >;
+            using base::base;
+
+            using base::lazy_into_block;
+            using base::iN;
+
+            logical_result matchAndRewrite(
```
xlauko, line 197: this would need a broader description what is happening here. Ideally example.

test/vast/Conversion/select-b.c
```diff
@@ -0,0 +1,35 @@
+// RUN: %vast-front -vast-emit-mlir=llvm -vast-snapshot-at="vast-core-to-llvm;vast-irs-to-llvm" %s
```
xlauko, line 1: Duplicit `RUN`s

xlauko (approved):

PR #593, opened 2024-05-28
Add runs without unsupported dialect to CI

Jezurko:
This also requires some modification to the benchmark runners themselves.

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -185,7 +185,7 @@ jobs:
       - name: Generate the results
         run: |
           python3 llvm-test-suite/utils/vast_compare.py \
-            --columns HighLevel with Unsupported,HighLevel,LLVM,Binary \
+            --columns HighLevel,"HighLevel without Unsupported",LLVM,Binary \
             --files results-hl-OFF.json,results-hl-ON.json,results-llvm-OFF.json,results-bin-OFF.json \
             --output single-source-results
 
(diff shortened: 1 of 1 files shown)


.github/workflows/prerelease.yml
```diff
@@ -270,8 +284,8 @@ jobs:
       - name: Generate the results
         run: |
           python3 sv-bench-patch/make_table.py \
-            --columns HighLevel \
-            --files results-sv-hl/results.txt \
+            --columns HighLevel,"HighLevel without Unsupported" \
+            --files results-sv-hl-false/results.txt,results-sv-hl-true/results.txt \
```
xlauko, line 288: Cannot we have a better suffix than `-false` and `-true`?

.github/workflows/prerelease.yml
```diff
@@ -177,8 +182,8 @@ jobs:
       - name: Generate the results
         run: |
           python3 llvm-test-suite/utils/vast_compare.py \
-            --columns vast-hl,vast-llvm,vast-bin \
-            --files results-hl.json,results-llvm.json,results-bin.json \
+            --columns HighLevel,"HighLevel without Unsupported",LLVM,Binary \
+            --files results-hl-OFF.json,results-hl-ON.json,results-llvm-OFF.json,results-bin-OFF.json \
```
xlauko, line 186: use better name than `-OFF` and `-ON`. Something like `results-hl-with-unsup`, `results-hl`.

.github/workflows/prerelease.yml
xlauko, line 1: You can add something like:

```

    - name: Set up results siffix
      run: |
        if [ "${{ matrix.disable_unsup }}" = "OFF" ]; then
          echo "VAST_UNSUP_SUFFIX=-with-unsup" >> $GITHUB_ENV
        fi
```

and use $VAST_UNSUP_SUFFIX later

xlauko (approved):

PR #599, opened 2024-05-29
abi: Support more floating point types.

lkorenc:
Add support for more cases when floating point types are passed by value.
Add relevant tests.
Refactor some older tests related to function calls.

### DIFF
--- a/include/vast/ABI/Classify.hpp
+++ b/include/vast/ABI/Classify.hpp
@@ -12,159 +12,132 @@ VAST_UNRELAX_WARNINGS
 
 #include "vast/Dialect/HighLevel/HighLevelUtils.hpp"
 
-namespace vast::abi
-{
-    template< typename H, typename ... Tail >
-    auto maybe_strip( mlir::Type t )
-    {
-        auto casted = [ & ]() -> mlir::Type
-        {
+namespace vast::abi {
+    template< typename H, typename... Tail >
+    auto maybe_strip(mlir::Type t) {
+        auto casted = [&]() -> mlir::Type {
             auto c = t.dyn_cast< H >();
-            if ( c )
+            if (c) {
                 return c.getElementType();
+            }
             return t;
         }();
 
-        if constexpr ( sizeof ... ( Tail ) == 0 )
+        if constexpr (sizeof...(Tail) == 0) {
             return casted;
-        else
-            return maybe_strip< Tail ... >( casted );
+        } else {
+            return maybe_strip< Tail... >(casted);
+        }
     }
 
     struct TypeConfig
     {
-        static bool is_void( mlir::Type t ) { return t.isa< mlir::NoneType >(); }
+        static bool is_void(mlir::Type t) { return t.isa< mlir::NoneType >(); }
 
-        static bool is_aggregate( mlir::Type t )
-        {
+        static bool is_aggregate(mlir::Type t) {
             // TODO(abi): `SubElementTypeInterface` is not good enough, since for
             //             example hl::PointerType implements it.
             // TODO(abi): Figure how to better handle this than manual listing.
-            return t.isa< hl::RecordType >()
-                   || t.isa< hl::ElaboratedType >()
-                   || t.isa< hl::ArrayType >();
+            return t.isa< hl::RecordType >() || t.isa< hl::ElaboratedType >()
+                || t.isa< hl::ArrayType >();
         }
 
-        static bool is_scalar( mlir::Type t )
-        {
+        static bool is_scalar(mlir::Type t) {
             // TODO(abi): Also complex is an option in ABI.
-            return !is_aggregate( t );
+            return !is_aggregate(t);
         }
 
-        static bool is_complex( mlir::Type t )
-        {
-            VAST_UNREACHABLE("");
-        }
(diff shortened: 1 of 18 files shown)


include/vast/ABI/Classify.hpp
```diff
@@ -177,14 +177,19 @@ namespace vast::abi
             return mlir::IntegerType::get( has_context.getContext(), s );
         }
 
+        static mlir_type fN(auto &mctx, std::size_t s) {
```
xlauko, line 180: add support for 16, 80 and 128

include/vast/ABI/Classify.hpp
```diff
@@ -232,6 +237,83 @@ namespace vast::abi
...
+                return {};
+            }();
+
+            // If we cannot get second type return first. It seems that `i8` would work
+            // as well.
+            if (!t1)
+                return t0;
+
+            // TODO: There is a bunch of conditions regarding size of `t0/t1` that
+            //       return `llvm::FIxedVector` types, which we currently cannot really
```
xlauko, line 290:
```suggestion
            //       return `llvm::FixedVector` types, which we currently cannot really
```

include/vast/ABI/Classify.hpp
```diff
@@ -232,6 +237,83 @@ namespace vast::abi
...
+        ){
+            if (offset == 0 && accept(t))
+                return t;
+
+            if (is_struct(t)) {
+                auto [field, field_start] = field_containing_offset(ctx, t, offset);
+                return type_at_offset(field, field_start, ctx, accept);
+            }
+
+            // array
```
xlauko, line 311: some TODO check or unreachable here?

include/vast/ABI/Classify.hpp
```diff
@@ -232,6 +237,83 @@ namespace vast::abi
...
+                auto nested = type_at_offset(t, offset + t0_size, ctx, get_is_sse_type());
+                if (t0_size > source_size && nested)
+                    return nested;
+
+                // We know `nested` was either not retrieved or that the sizes are not
+                // good enough.
+
+                // Check case of half/bfloat + float.
+                if (size(dl, t0) == 16 && source_size > 4) {
+                    // `+4` comes from alignement
```
xlauko, line 278: is this alignement fixed? if it can be altered from outside make it defensive to catch such cases.

lkorenc: It is hardcoded in clang as well so I doubt it can be changed as this is result of how underlying arch works (other archs will do this completely differently).

include/vast/ABI/Classify.hpp
```diff
@@ -232,6 +243,86 @@ namespace vast::abi
...
+        }
+
+        // TODO: In `clang` int and sse variants have separate implementations (there is
+        //       a lot of difference) but in the core they are doing very similar things.
+        //       It should be possible to make a shared helper that takes away a lot of
+        //       shared complexity.
+        static mlir_type sse_target_type_at_offset(
+            mlir_type t, std::size_t offset,
+            mlir_type root, std::size_t root_offset,
+            const auto &ctx
```
xlauko, line 259:
```suggestion
            const auto &abi_ctx
```

include/vast/ABI/Classify.hpp
```diff
@@ -232,6 +243,86 @@ namespace vast::abi
             return mlir::IntegerType::get( t.getContext(), final_size );
         }
 
+        static auto get_is_sse_type() {
+            return [](auto t) {
```
xlauko, line 247:
```suggestion
            return [](mlir_type t) {
```

include/vast/ABI/Classify.hpp
xlauko: In general, it will improve readability if you add named type, e.g. `abi_ctx_t` and use that instead of auto if possible. Also use `abi_ctx` everywhere where applicable, as we have various contexts and one need to backtrack where these originates from.

include/vast/ABI/Classify.hpp
```diff
@@ -566,10 +657,11 @@ namespace vast::abi
                 if ( hi_start == 8 )
                     return lo;
                 // TODO(abi): float, half -> promote to double
-                VAST_CHECK( TypeConfig::is_scalar_integer( lo ) ||
-                            TypeConfig::is_pointer( lo ),
-                            "Cannot adjust half type for pair {0}. {1}", lo, hi );
-                return TypeConfig::iN( lo, 64 );
+                if (TypeConfig::is_scalar_integer( lo ) || TypeConfig::is_pointer( lo ))
```
xlauko, line 660:
```suggestion
                if (TypeConfig::is_scalar_integer(lo) || TypeConfig::is_pointer(lo))
```

xlauko: Literally right next condition uses different style

include/vast/ABI/Classify.hpp
lkorenc: Fair enough, although this is not exactly `abi_ctx`, but I will come up with a good name for it.

include/vast/ABI/Classify.hpp
lkorenc: Will include a reformat commit at the end of the PR.

PR #611, opened 2024-06-07
abi: Support array types.

lkorenc:
Adds support for array types and includes various fixes that were implemented along the way.


include/vast/ABI/Classify.hpp
```diff
@@ -756,8 +571,7 @@ namespace vast::abi {
                 }
                 case Class::SSE: {
                     ++needed_sse;
-                    auto target_type =
-                        TypeConfig::sse_target_type_at_offset(t, 8, t, 8, mk_classifier_ctx());
+                    auto target_type = sse_target_type_at_offset(t, 64, t, 64);
```
xlauko, line 574: 8 changed to 64. Was it intended?

include/vast/ABI/MLIRTypeInfo.hpp
```diff
@@ -0,0 +1,252 @@
...
+#include <llvm/Support/Alignment.h>
+VAST_UNRELAX_WARNINGS
+
+#include "vast/ABI/ABI.hpp"
+
+#include "vast/Dialect/HighLevel/HighLevelUtils.hpp"
+
+namespace vast::abi {
+
+    struct vast_type_info {
```
xlauko, line 17: does it need to be called with prefix `vast_` as it is already in `vast::abi`

include/vast/ABI/MLIRTypeInfo.hpp
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    bool vast_type_info::is_complex(mlir_type t) {
+        VAST_UNREACHABLE("Missing support for complex types!");
+    }
+
+    // TODO(abi): `SubElementTypeInterface` is not good enough, since for
+    //             example hl::PointerType implements it.
+    // TODO(abi): Figure how to better handle this than manual listing.
+    bool vast_type_info::is_aggregate(mlir_type t) {
```
xlauko, line 95: have custom `AggregateTypeInterface`

include/vast/ABI/MLIRTypeInfo.hpp
```diff
@@ -0,0 +1,252 @@
...
+            return true;
+        }
+        if (auto et = t.dyn_cast< hl::ElaboratedType >()) {
+            return is_record(et.getElementType());
+        }
+        return false;
+    }
+
+    bool vast_type_info::is_struct(mlir_type t) {
+        // TODO(abi): Are these equivalent?
```
xlauko, line 110: yes keep only one

include/vast/ABI/MLIRTypeInfo.hpp
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    // TODO(abi): `SubElementTypeInterface` is not good enough, since for
+    //             example hl::PointerType implements it.
+    // TODO(abi): Figure how to better handle this than manual listing.
+    bool vast_type_info::is_aggregate(mlir_type t) {
+        return mlir::isa< hl::RecordType, hl::ElaboratedType, hl::ArrayType >(t);
+    }
+
+    bool vast_type_info::is_record(mlir_type t) {
```
xlauko, line 99: just `return mlir::isa< hl::RecordType >(hl::strip_elaborated(t));`

include/vast/ABI/MLIRTypeInfo.hpp
```diff
@@ -0,0 +1,252 @@
...
+        return { array_type.getSize(), array_type.getElementType() };
+    }
+
+    bool vast_type_info::can_be_passed_in_regs(mlir_type t) {
+        // TODO(abi): Seems like in C nothing can prevent this.
+        return true;
+    }
+
+    bool vast_type_info::is_scalar_integer(mlir_type t) {
+        return t.isa< mlir::IntegerType >();
```
xlauko, line 131:
```suggestion
        return mlir::isa< mlir::IntegerType >(t);
```

include/vast/ABI/MLIRTypeInfo.hpp
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    // TODO(abi): Implement.
+    bool vast_type_info::has_unaligned_field(mlir_type t) {
+        return false;
+    }
+
+    bool vast_type_info::is_scalar_float(mlir_type t) {
+        return t.isa< mlir::FloatType >();
```
xlauko, line 140:
```suggestion
        return mlir::isa< mlir::FloatType >(t);
```

xlauko: what about `hl` float types?

include/vast/ABI/MLIRTypeInfo.hpp
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    bool vast_type_info::is_pointer(mlir_type t) {
+        return t.isa< hl::PointerType >();
```
xlauko, line 144:
```suggestion
        return mlir::isa< hl::PointerType >(t);
```

include/vast/ABI/MLIRTypeInfo.hpp
```diff
@@ -0,0 +1,252 @@
...
+            auto field_info = field_containing_offset(t, offset);
+            if (!field_info)
+                return {};
+            auto [field, field_start] = *field_info;
+            return type_at_offset(field, field_start, accept);
+        }
+
+        if (is_array(t)) {
+            auto [_, element_type] = array_info(t);
+            auto element_size = size(element_type);
```
xlauko, line 222:
```suggestion
            const auto element_size = size(element_type);
```

include/vast/ABI/MLIRTypeInfo.hpp
```diff
@@ -0,0 +1,252 @@
...
+            auto element_offset = offset / element_size * element_size;
```
xlauko, line 223:
```suggestion
            const auto element_offset = offset / element_size * element_size;
```

include/vast/ABI/MLIRTypeInfo.hpp
lkorenc: Probably not. Will remove the prefix.

include/vast/ABI/MLIRTypeInfo.hpp
lkorenc: I will adjust/rework the comment - as this follows the clang codebase this will be expanded once classes are in the picture.

include/vast/ABI/MLIRTypeInfo.hpp
lkorenc: Already not present in this layer in the VAST conversion pipelines. Technically we probably should support it, but it is missing at more than just this place. I will open an issue - ideally we would want to have a separate class for just VAST-only types and one for upstream types.

include/vast/ABI/Classify.hpp
lkorenc: Yes, before it was bug because we were sending in size in bytes (as clang does) but our helpers return sizes in bits. So I unified it and used bits all over the place. This will really benefit from having strong types for bit/byte sizes, there is already an open issue for that.

include/vast/ABI/MLIRTypeInfo.hpp
lkorenc: Opening issue for it.

include/vast/ABI/Classify.hpp
```diff
@@ -13,242 +13,135 @@ VAST_UNRELAX_WARNINGS
...
+            if (type_info.size(t) <= start) {
                 return true;
             }
 
-            if (is_array(t)) {
-                VAST_TODO("bits_contain_no_user_data, array: {0}", t);
-            }
-
-            if (is_record(t)) {
+            if (type_info.is_record(t) || type_info.is_array(t)) {
```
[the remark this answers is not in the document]
lkorenc: Yeah for now why not.

xlauko (approved):

Issue #645, opened 2024-07-02
[Bug]: Verifier fails to match binary operand types when one of them is a typeof type


PappasBrent:
### VAST version

Ubuntu clang version 18.1.8 (++20240615103753+3b5b5c1ec4a3-1~exp1~20240615223858.136) Target: x86_64-pc-linux-gnu Thread model: posix

### LLVM version

18.1.8

### Operating system

Ubuntu 22.04.4 LTS x86_64

### Description

The VAST codegen module verifier fails to verify binary expressions when one of the operands is declared using a `typeof` expression to decide its type. I receive the following error:

```
loc("test_bin.c":3:14): error: 'hl.add' op failed to verify that is an additive operation (types match or are ptr and integral)
[VAST fatal] codegen: module verification error before running vast passes
PLEASE submit a bug report to https://github.com/trailofbits/vast/issues and include the crash backtrace, preprocessed source, and associated run script.
Stack dump:
0.      Program arguments: ./builds/default/tools/vast-front/Debug/vast-front -x c -vast-emit-mlir=hl test_bin.c
1.      <eof> parser at end of file
 #0 0x00007f0838542716 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) (/usr/lib/llvm-18/lib/libLLVM.so.18.1+0xd94716)
 #1 0x00007f08385406d0 llvm::sys::RunSignalHandlers() (/usr/lib/llvm-18/lib/libLLVM.so.18.1+0xd926d0)
 #2 0x00007f08384921f0 (/usr/lib/llvm-18/lib/libLLVM.so.18.1+0xce41f0)
 #3 0x00007f0837230520 (/lib/x86_64-linux-gnu/libc.so.6+0x42520)
 #4 0x000056359c52ca13 vast::cg::driver::finalize() /home/bpappas/github.com/trailofbits/vast/lib/vast/CodeGen/CodeGenDriver.cpp:76:17
 #5 0x000056359c524166 vast::cc::vast_consumer::HandleTranslationUnit(clang::ASTContext&) /home/bpappas/github.com/trailofbits/vast/lib/vast/Frontend/Consumer.cpp:67:5
 #6 0x000056359c5249e7 vast::cc::vast_stream_consumer::HandleTranslationUnit(clang::ASTContext&) /home/bpappas/github.com/trailofbits/vast/lib/vast/Frontend/Consumer.cpp:125:15
[...]

PR #658, opened 2024-07-09
Modify type verifying utils to work with typeof types

Jezurko:
#645


Jezurko:
@PappasBrent this should hopefully fix the `typeof` related issues, if I haven't missed any corner cases. I almost missed the `float`s /o\ :D

PappasBrent:
Testing locally it seems to fix the issue with shift operations that I reported earlier today; I'll test it some more to be sure (and run it on the Linux kernel as well)

lkorenc (approved):

xlauko (approved):

PR #660, opened 2024-07-10
Core symbol tables

xlauko:


include/vast/Conversion/ABI/AggregateTypes.hpp
```diff
@@ -179,7 +179,7 @@ namespace vast::conv::abi {
         };
 
         state_t &state;
-        vast_module mod;
+        core::module mod;
```
lkorenc, line 182: Why is `module_op_t` not used here (and at other places in this PR as well)?

Jezurko (approved):

PR #671, opened 2024-07-12
Add steps to run vast-front on Linux kernel

PappasBrent:
Add steps to download, configure, and build the Linux kernel; run vast-front on its compilation units; and tabulate the results.

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -413,6 +413,7 @@ jobs:
         with:
           pattern: vast_linux_kernel_times_*
           merge-multiple: true
+
       - name: Install converter dependencies
         run: pip3 install pandas scipy tabulate
 
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -336,10 +336,13 @@ jobs:
      - name: Export vast binaries
        run: echo "${PWD}/vast/bin/" >> $GITHUB_PATH
 
-     - name: Install benchmark dependencies
-       run: |
+    - name: Install benchmark dependencies
+      run: |
           apt-get update
-          apt-get -y install bear
+          apt-get -y install git fakeroot build-essential \
+            ncurses-dev xz-utils libssl-dev bc flex libelf-dev bison \
+            bear \
+            apt-get -y install lld-${{ matrix.llvm-version }}
 
      - name: Clone the Linux kernel and build it to create compiliation database
        run: |
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -355,11 +355,11 @@ jobs:
      - name: Clone vast benchmark directory
        uses: actions/checkout@v4
        with:
-         repository: trailofbits/vast-benchmarks
-         sparse-checkout: benchmarks/linux_kernel
-         ref: main
-         path: .
-         fetch-depth: 1
+          repository: trailofbits/vast-benchmarks
+          sparse-checkout: benchmarks/linux_kernel
+          ref: linux-benchmark-make-output-optional
+          path: vast-benchmarks/
+          fetch-depth: 1
 
      - name: Run benchmarks
        run: >
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -447,7 +447,7 @@ jobs:
         image-version: [22.04]
     name: "Build VAST doc"
     runs-on: ubuntu-${{ matrix.image-version }}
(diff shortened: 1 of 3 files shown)


.github/workflows/prerelease.yml
```diff
@@ -304,6 +304,59 @@ jobs:
...
+         && bear -- make LLVM=1
+
+     - name: Download the benchmark runner
+       run: wget https://raw.githubusercontent.com/PappasBrent/vast-benchmarks/add-linux-kernel-benchmark/benchmarks/linux_kernel/run_vast_benchmark.py -O run_vast_benchmark.py
+
+     - name: Run benchmarks
+       run: >
+         python3 run_vast_benchmark.py
+         vast-front # TODO: Determine if this is the correct path to vast-front
+         linux/compile_commands.json
```
Jezurko, line 353: `vast-front` should be in path after the export step, so it should be fine

.github/workflows/prerelease.yml
```diff
@@ -304,6 +304,59 @@ jobs:
...
+
+     - name: Clone the Linux kernel and build it to create compiliation database
+       run: >
+         git clone --depth=1 https://github.com/torvalds/linux.git linux
+         && cd linux
+         && make defconfig
+         && bear -- make LLVM=1
+
+     - name: Download the benchmark runner
+       run: wget https://raw.githubusercontent.com/PappasBrent/vast-benchmarks/add-linux-kernel-benchmark/benchmarks/linux_kernel/run_vast_benchmark.py -O run_vast_benchmark.py
```
Jezurko, line 347: You can use the `checkout` action with `sparse-checkout` option to get a single file from a github repository. (see `eval_llvm_ts` job)

.github/workflows/prerelease.yml
```diff
@@ -304,6 +304,59 @@ jobs:
...
+
+     - name: Run benchmarks
+       run: >
+         python3 run_vast_benchmark.py
+         vast-front # TODO: Determine if this is the correct path to vast-front
+         linux/compile_commands.json
+         vast_linux_kernel_mlir_with_unsup/
+         --num_processes=8
+         --vast_option="-xc"
+         --vast_option="-vast-emit-mlir=hl"
```
Jezurko, line 357: Take the mlir target from the `vast-target` job variable, so that we can configure it to run into other dialects.

.github/workflows/prerelease.yml
```diff
@@ -304,6 +304,59 @@ jobs:
...
+
+     - name: Export vast binaries
+       run: echo "${PWD}/vast/bin/" >> $GITHUB_PATH
+
+     - name: Clone the Linux kernel and build it to create compiliation database
+       run: >
+         git clone --depth=1 https://github.com/torvalds/linux.git linux
+         && cd linux
+         && make defconfig
+         && bear -- make LLVM=1
```
Jezurko, line 344: You will probably need to install `bear` before running.
Also you can use the run directive with `|` for multicommand `run` step ([](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun))

.github/workflows/prerelease.yml
PappasBrent: Ah cool

.github/workflows/prerelease.yml
PappasBrent: Oh very neat

PappasBrent:
This PR adds CI jobs to run `vast-front` on all files in the default configuration of the Linux kernel and combine the timing results into a Markdown table, but does not change the `build_doc` job to add the generated table to VAST's website. Should this PR handle this step as well or should I open a separate PR for this?

.github/workflows/prerelease.yml
```diff
@@ -304,6 +304,139 @@ jobs:
...
+      - name: Run benchmarks without unsupported
+        if: ${{ matrix.disable-unsup }}
+        run: >
+          python3 ${PWD}/vast-benchmarks/benchmarks/linux_kernel/run_vast_benchmark.py
+          vast-front
+          ${PWD}/linux/compile_commands.json
+          --num_processes=$(nproc)
+          --vast_option="-xc"
+          --vast_option="-vast-emit-mlir=${{ matrix.vast-target }}"
+          --vast_option="-vast-disable-unsupported"
```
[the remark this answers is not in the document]
PappasBrent: I was actually looking for a way to do this; thanks for the advice.

Jezurko (approved):

PR #683, opened 2024-07-29
Linkage specials

Jezurko:
Fixup linkage in cases where clang has annoying assertions
Open to suggestions how to resolve this in a better way


lib/vast/CodeGen/CodeGenFunction.cpp
```diff
@@ -47,23 +47,60 @@ namespace vast::cg
     }
 
     operation function_generator::emit(const clang_function *decl) {
+        decl->dump();
```
xlauko, line 50: dump?

lib/vast/CodeGen/CodeGenFunction.cpp
```diff
@@ -47,23 +47,60 @@ namespace vast::cg
...
+                        "Mismatch in number of arguments between function declaration and "
+                        "vast function."
+                    );
+                    if (fn_args == 0 && decl_args != 0) {
+                        fn.setType(visitor.visit(decl->getFunctionType()));
+                    }
+
+                    // inline function declarations (without definition) are unknown
+                    // because clang can not provide the information for them
+                    auto linkage = fn->getAttrOfType< core::GlobalLinkageKindAttr >("linkage");
```
xlauko, line 71: this duplicates code ... make a function that takes desired linkage, i.e., core::get_function_linkage(decl) or core::GlobalLinkageKind::ExternalLinkage and sets linkage attr and visibility

xlauko (approved):

PR #685, opened 2024-07-31
Generate Linux compile commands, don't intercept with bear

PappasBrent:
- Generate Linux compile commands with the `gen_compile_commands.py` script instead of using `bear` to do so since `bear`'s output does not handle single quotes in command-line macro definitions.
- Add `make prepare` step when building the Linux kernel.
- Checkout Linux v6.10 specifically to improve reproducibility.

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -347,7 +347,7 @@ jobs:
         run: |
           git clone --depth=1 https://github.com/torvalds/linux.git linux
           cd linux
-          make defconfig
+          make defconfig LLVM=-${{ matrix.llvm-version }}
           make LLVM=-${{ matrix.llvm-version }} -j $(nproc)
           python3 ./scripts/clang-tools/gen_compile_commands.py
           cd ../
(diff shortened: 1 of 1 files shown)


.github/workflows/prerelease.yml
```diff
@@ -341,15 +337,16 @@ jobs:
...
-          git clone --depth=1 https://github.com/torvalds/linux.git linux
+          git clone https://github.com/torvalds/linux.git --depth=1 --branch=v6.10 linux
           cd linux
-          bear -- make defconfig
-          bear -- make LLVM=-${{ matrix.llvm-version }} -j $(nproc)
+          make defconfig LLVM=-${{ matrix.llvm-version }}
+          make prepare LLVM=-${{ matrix.llvm-version }}
+          make LLVM=-${{ matrix.llvm-version }} -j $(nproc)
+          python3 ./scripts/clang-tools/gen_compile_commands.py
           cd ../
```
Jezurko, line 350: Sorry, I missed this before. I believe you do not need this `cd ../` command, as the next step again should start in the "original" working directory (or in a directory set by the `working-directory` option)

Jezurko (approved):

PR #697, opened 2024-08-09
Bump release action version in CI

Jezurko:
Also should fix improperly generated release notes

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -65,7 +65,7 @@ jobs:
           retention-days: 1
 
       - name: Publish Pre-Release
-        uses: softprops/action-gh-release@v1
+        uses: softprops/action-gh-release@v2
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/prerelease.yml
```diff
@@ -65,9 +65,9 @@ jobs:
           retention-days: 1
 
       - name: Publish Pre-Release
-        uses: softprops/action-gh-release@v1
+        uses: softprops/action-gh-release@v2
         with:
-          tag_name: "latest"
+          name: "Pre-Release"
```
xlauko, line 70: why did you removed tag_name?

Jezurko: because `latest` is some old tag that nobody is updating. If I understood the action documentation properly, the variable should be default initialized to the branch to which was pushed (which should always be `master` for this action)

xlauko (approved):

PR #712, opened 2024-08-30
Expand support for attributes found in linux kernel

Jezurko:


include/vast/Dialect/HighLevel/HighLevelAttributes.td
```diff
@@ -31,11 +31,47 @@ class HighLevel_NameAttr<string name, string attr_mnemonic>
   let assemblyFormat = "`<` $name `>`";
 }
 
+class HighLevel_SymbolReferringAttr<string name, string attr_mnemonic>
```
xlauko, line 34: This should be core attr as it does not have direct `AST` equivalent and can be useful elsewhere.

xlauko (approved):

PR #730, opened 2024-09-23
Add support for cells with non-local storage

Jezurko:
- [x] Move storage attributes from `hl` to `core`
- [x] Add Interface for storage information on declarations
- [x] Modify cell declarations to keep storage information
- [x] Nuke GlobalRefs and replace them with DeclRefs
- [x] Lower non-local cells to llvm 
- [x] Add tests


test/vast/Dialect/HighLevel/glob-linkage-a.c
```diff
@@ -0,0 +1,16 @@
...
+// CHECK: hl.var @wundef, <weak>
+int __attribute__((weak)) wundef;
+// CHECK: hl.var @wdef, <weak>
+int __attribute__((weak)) wdef = 5;
+// CHECK: hl.var @ewdef, <weak>
+extern int __attribute__((weak)) ewdef = 5;
+// CHECK: hl.var @edef, <external>
+extern int edef = 5;
+// CHECK: hl.var @undef, <common>
+int undef;
```
Jezurko, line 14: From my understanding this is correct (see [clang/llvm](https://llvm.org/docs/LangRef.html#linkage-types) documentation), but when testing this in [godbolt](https://godbolt.org/z/KzKTKzre8) or locally it actually has `external` linkage. Not sure how to proceed with this

lib/vast/Conversion/FromHL/Passes.cpp
```diff
@@ -42,6 +42,10 @@ namespace vast::conv::pipeline {
         return pass(createVarsToCellsPass);
     }
 
+    pipeline_step_ptr evict_static_locals() {
+        return pass(createEvictStaticLocalsPass);
```
xlauko, line 46: add `.depends_on(refs_to_ssa)`?

test/vast/Conversion/subscript-b.c
```diff
@@ -2,25 +2,25 @@
...
+// STD_TYPES: hl.var @arr1, <external> : !hl.lvalue<!hl.array<3, si32>> = {
 
 
-// VAL_CAT: hl.var @arr1 : !hl.ptr<!hl.array<3, si32>> = {
+// VAL_CAT: hl.var @arr1, <external> : !hl.ptr<!hl.array<3, si32>> = {
 // VAL_CAT:    hl.value.yield {{.*}} : !hl.array<3, si32>
 
 
-// C_LLVM:  llvm.mlir.global internal constant @arr1() {addr_space = 0 : i32} : !llvm.array<3 x i32> {
+// C_LLVM:  llvm.mlir.global external @arr1() {addr_space = 0 : i32} : !llvm.array<3 x i32> {
```
xlauko, line 12: <3

xlauko (approved):

lib/vast/Conversion/FromHL/Passes.cpp
Jezurko: I don't think it depends on `refs_to_ssa`. I believe these passes can be run in any order, as the legalizers should make sure that the pass does not act on the wrong Variable declarations

PR #738, opened 2024-10-17
Fix enum lowering of shadowed values

xlauko:


xlauko:
- introduces scope aware type conversion
- adds empty symbol table trait
- refactor enum lowering passes

Jezurko (approved):

PR #750, opened 2024-11-22
Parser type conversions

xlauko:


include/vast/Conversion/TypeConverters/TypeConverter.hpp
```diff
@@ -45,6 +45,19 @@ namespace vast::conv::tc {
         }
     };
 
+    static inline constexpr auto unrealized_materialization = [] (
```
frabert, line 48: Why a lambda and not a regular static function?

include/vast/Conversion/TypeConverters/TypeConverter.hpp
```diff
@@ -139,6 +152,10 @@ namespace vast::conv::tc {
                 .template take_wrapped< maybe_type_t >();
         }
 
+        maybe_type_t convert_arg_type(mlir_type t, unsigned long /* idx */) const {
```
frabert, line 155: I see that `convert_arg_type` here takes an additional parameter `idx` which is currently unused, but forces us to iterate over arguments using `llvm::enumerate` later on. Are we sure we'll need the `idx` parameter in the future?

tools/detect-parsers/main.cpp
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) 2021-present, Trail of Bits, Inc.
```
frabert, line 1: This file is probably old

frabert (changes requested):

include/vast/Conversion/TypeConverters/TypeConverter.hpp
xlauko: It is there as an extension point, there are overloads that use the index.

frabert (approved):

PR #758, opened 2025-01-10
Introduce typed symbol references

xlauko:
- adds typed references
- implements type aware symbol resolution

### DIFF
--- a/include/vast/Dialect/HighLevel/HighLevelOps.td
+++ b/include/vast/Dialect/HighLevel/HighLevelOps.td
@@ -339,7 +339,7 @@ def HighLevel_FieldDeclOp
 
 def HighLevel_RecordMemberOp
   : HighLevel_Op< "member" >
-  , Arguments<(ins AnyType:$record, FlatSymbolRefAttr:$field)>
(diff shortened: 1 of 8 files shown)


include/vast/Dialect/HighLevel/HighLevelOps.td
```diff
@@ -339,7 +339,7 @@ def HighLevel_FieldDeclOp
 
 def HighLevel_RecordMemberOp
   : HighLevel_Op< "member" >
-  , Arguments<(ins AnyType:$record, FlatSymbolRefAttr:$field)>
+  , Arguments<(ins AnyType:$record, Core_MemberVarSymbolRefAttr:$field)>
```
Jezurko, line 342: Will this work properly with function pointers inside structs?

xlauko: huh? Core_MemberVarSymbolRefAttr is just alias for FlatSymbolRefAttr.

Jezurko: Nevermind, yeah. At first I thought it might break the lookup when a call of a function happens using record member, but it was just a brainfart.

PR #760, opened 2025-01-14
VarDecl to Parser lowering

xlauko:


lib/vast/Conversion/Parser/ToParser.cpp
```diff
@@ -493,13 +494,97 @@ namespace vast::conv {
...
+
+                rewriter.inlineBlockBefore(body, op);
+                rewriter.replaceOp(op, yield.op().getResult());
+                rewriter.eraseOp(yield.op());
+
+                return mlir::success();
+            }
+
+            static void legalize(parser_conversion_config &cfg) {
+                cfg.target.addLegalOp< mlir::UnrealizedConversionCastOp >();
```
Jezurko, line 525: Do we have to legalize it here? Shouldn't that happen automatically?

Jezurko (approved):

PR #762, opened 2025-01-15
Unary parser operations

xlauko:


lib/vast/Conversion/Parser/ToParser.cpp
```diff
@@ -233,9 +233,10 @@ namespace vast::conv {
         }
 
         using value_range = mlir::ValueRange;
+        using type_range = mlir::TypeRange;
```
Jezurko, line 236: Move these to global utils?

Jezurko (approved):

PR #780, opened 2025-04-02
Refactor communication and model parsing for AI integration

frabert:


xlauko (approved):

frabert:
The AI-provided function models are very likely not 100% accurate -- for example I noticed the LLM tends to err on the side of marking something as a `sink` instead of `noparser` when it wasn't sure, but I think they are more useful than not having them at all.

